PR #4, opened 2014-03-12
Fixed include tags and added inertial tags

sepjansen:
We noticed two issues with the robotiq model:
1. In hydro, including other xacro files using <include> tag were deprecated. Changed all of these to xacro:include
2. Model could not be used in gazebo, since no inertial information was modeled. I added inertial information to all links of the articulated model, based on CAD models and some assumptions.


shaun-edwards:
+1

PR #5, opened 2014-04-14
Adding USB Modbus support and RobotiqCGripper

kphawkins:
This adds nodes and drivers which allow you to interface with the USB Modbus interface for the gripper. Furthermore, it adds a Python utility class which interfaces with the current topics and controls basic gripper functionality.


nicolaslauzier:
I have looked at the code and I really like the modifications. Especially the "robotiq_c_ctrl" package which adds easier functions to use than the control variables that we describe in our documentation. I have a few comments stated below. If you are willing to make these modifications, that would be nice. Otherwise, we can accept the code as is and modify it later. It's mainly a matter of being coherent with the nomenclature and descriptions around the code itself:
- There is no such thing as Modbus USB. I would prefer to name it Modbus RTU since the code that you added will be valid for all serial interfaces that we provide (RS-232, RS-485, USB). We do not advertise this yet, but the last version of the 3-Finger model can be controlled via a RS-485 connection that passes through the power cable. Also, the 2F-200 model (the big one) can be controlled via RS-485 directly (without the black box) and we would like to extend that to the small gripper in a near future. The control interface will be the same so you code will run "as is" on these models, which is great!
- It would be preferable to create another package for Modbus RTU or at least rename "comModbusTCP" for "comModbus" in order to avoid weird things such as "gripper.client = robotiq_modbus_tcp.comModbusUsb.communication()".
  
  In all cases, I think the contribution is very nice and I'm really happy that this functionality was added. I just want to make sure that it will fit well as our products evolve.

Regards,

Nicolas, Robotiq inc.

shaun-edwards:
@kphawkins, is this something you could do?  If not I propose we merge the request and then create isses for @nicolaslauzier comments.

kphawkins:
Sorry for not replying sooner. It's definitely something I can do, but it
might be ~1 week before I get around to making the changes and testing it
on the actual gripper.  Do whatever seems best.

-Kelsey

nicolaslauzier:
Hi Kelsey,

Thanks for accepting the modification requests. The time frame you mention is perfect and I certainly do not want to put any pressure on you. It's a very nice contribution and if we can do anything, please do not hesitate.

Also, we do plan a better integration of our gripper on the UR robots. If you have any suggestions or comments, I would be happy to hear them.

Regards,

Nicolas

shaun-edwards:
@kphawkins, ping.

kphawkins:
I have renamed the packages/variables as requested and tested it on my platform.  I opted to split the modbus package into two separate packages to avoid conflicting with existing implementations.  @shaun-edwards can we eventually merge the two implementations into a single `robotiq_modbus` package with minimal overlapping code?

As for gripper integration for UR robots, I'd love to be able to run the C-model's cabling through the UR10's existing cable outlet.  I understand that current limitations might be a problem, but I don't mind trading reduced gripper power for internal cables.

nicolaslauzier:
Thanks Kelsey for these changes. I think splitting into two packages makes sense.

I agree that all end-effector integration should not have any external cable. We are discussing with UR to do use their internal cabling but this will not exist this year. The problem is that they use the communication bus for their actuators so the bandwidth is already quite used and they would not like to have uncontrolled messages on the bus. We'll have to see if they add another mean of communicating with the tool beside their IO board they currently offer. 

Thanks for the info though. If we come up with it, we know that it will be appreciated! =o)

PR #6, opened 2014-05-14
Added type check to response in getStatus

orikuma:
When I use baseSModel.py and execute sendCommand to robotiq s model in short time, getStatus failes by following error.
I could avoid this error by this changes, but I think it is not good that getStatus has while loop in itself and I want to know why read_input_registers returns WriteMultipleRegistersResponse.
Is this problem caused by RobotiqHand firmware or pymodbus?

```
File "/home/leus/ros/hydro/src/ros-industrial/robotiq_s_model_control/src/robotiq_s_model_control/baseSModel.py", line 155, in getStatus
    status = self.client.getStatus(16);
File "/home/leus/ros/hydro/src/ros-industrial/robotiq_modbus_tcp/src/robotiq_modbus_tcp/comModbusTcp.py", line 89, in getStatus
    output.append((response.getRegister(i) & 0xFF00) >> 8)
AttributeError: 'WriteMultipleRegistersResponse' object has no attribute 'getRegister'
```

### DIFF
--- a/robotiq_modbus_tcp/src/robotiq_modbus_tcp/comModbusTcp.py
+++ b/robotiq_modbus_tcp/src/robotiq_modbus_tcp/comModbusTcp.py
@@ -45,11 +45,13 @@ from pymodbus.client.sync import ModbusTcpClient
 from pymodbus.register_read_message import ReadInputRegistersResponse
 from math import ceil
 import time
+import threading
 
-class communication:	
+class communication:
 
    def __init__(self):
       self.client = None
+      self.lock = threading.Lock()
       
    def connectToDevice(self, address):
       """Connection to the client - the method takes the IP address (as a string, e.g. '192.168.1.11') as an argument."""
@@ -72,8 +74,9 @@ class communication:
       for i in range(0, len(data)/2):
          message.append((data[2*i] << 8) + data[2*i+1])
 
-      #To do!: Implement try/except 
-      self.client.write_registers(0, message)
+      #To do!: Implement try/except
+      with self.lock:
+         self.client.write_registers(0, message)
 
    def getStatus(self, numBytes):
       """Sends a request to read, wait for the response and returns the Gripper status. The method gets the number of bytes to read as an argument"""
@@ -81,10 +84,8 @@ class communication:
 
       #To do!: Implement try/except 
       #Get status from the device
-      response = None
-      while not isinstance(response, ReadInputRegistersResponse):
+      with self.lock:
          response = self.client.read_input_registers(0, numRegs)
-         time.sleep(0.1)
 
       #Instantiate output as an empty list
       output = []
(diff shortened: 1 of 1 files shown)


shaun-edwards:
@nicolaslauzier, Do you have comment on this?

nicolaslauzier:
The problem is with pymodbus, or its use. The probable cause is that the request was that the request for "WriteMultipleRegistersResponse" was sent but the answer was not received in time due to some delay on the network (or error in the messages). A new request was sent but the echo message (acknowledgment) for the write request was received after the new request was sent.

Checking the type of the answer is a working solution but I think it would be better to wait for the echo message to be received before sending a new request. Can you take a look at this solution? I can help if you are not familiar with pymodbus.

shaun-edwards:
@orikuma, ping

orikuma:
I'm sorry for the delay in contact.
This problem was caused by my use in sending requests.
I used sendCommand() in ros subscriber callback function and getStatus() in main loop.
When a read request was sent by getStatus in main loop immediately after a write request was sent by sendCommand in callback, response in getStatus receive response to write request and reported error occured.
Therfore, I think the correct solution for this problem is introducing exclusive control like threading.Lock before sending request by sendCommand, getStatus etc in user functions.
Thank you for your kind support.

shaun-edwards:
@orikuma, Thank you for the quick reply and the update.  It sounds like this should be logged as an issue against the robotiq package.  I'm not sure that the robotiq package was written with thread safety in mind, but adding it would avoid this issue.  I'd like to understand how critical this issue is.  Have you been able to work around the issue or is it preventing you from utilizing the robotiq library?

By the way, I really appreciate you using the robotiq libraries and offering up feedback when it doesn't meet you needs.

orikuma:
I could noticed the cause of this problem and deal with it by your advice to wait for the echo message to be received before sending a new request.
Now, I have been able to work around the problem and I am using the robotiq library without problems.

I think this problem is critical in use because comModbusTcp is terminated forcely by this error.
In addition, I think this problem likely to happen in other user programs because sendCommand is probably used in subscriber callback.
I cannot have the conviction whether thread safety should be guaranteed in comModbusTcp (ex. add threading.Lock to communication class in comModbusTcp) or in user functions, but It would be helpful to let user know whether the robotiq library considers thread safety or not.

nicolaslauzier:
Is your workaround part of SModelTcpNode.py? Do you think you can commit it to the repo so that we can make it part of the library?

Thanks!

orikuma:
My workaround is for communication class in comModbusTcp.py, not SModelTcpNode.py.
I simply added threading.Lock() exclusive control to communication class to avoid reported error.
Please check the changes. Thanks.

shaun-edwards:
@nicolaslauzier , It looks like @orikuma , has update his PR.  Can you review?  Thanks in advance!

nicolaslauzier:
The modification is very nice. It's indeed the proper way of implementing it (it is way better than having delays). I should have done that from the beginning but I wasn't experiencing the issue and I'm not familiar with threading in Python.

I think that we should merge this to the trunk. @shaun-edwards, can you do that?

Thanks @orikuma  for this modification!

Nicolas

shaun-edwards:
Thanks @orikuma & @nicolaslauzier

Issue #2, opened 2014-02-06
Deprecated include tag in hydro


sepjansen:
When trying to convert to xacro models to urdf, the following error occurs:

```
DEPRECATED IN HYDRO:
  The <include> tag should be prepended with 'xacro' if that is the intended use 
  of it, such as <xacro:include ...>. Use the following script to fix incorrect
  xacro includes:
     sed -i 's/<include/<xacro:include/g' `find . -iname *.xacro`
```

fevb:
I would suggest the developers to open a 'hydro-devel' branch so that we can submit a pull request to fix that issue

shaun-edwards:
@fevb, I've created the branch.  Thanks for the suggestion.

PR #8, opened 2014-06-27
changed deprecated include tag to xacro:include in .xacro files

fevb:
changed deprecated include tag to xacro:include in .xacro files in visualization package


shaun-edwards:
This changed addressed #2

PR #13, opened 2014-12-04
added gazebo launch files

UltronDestroyer:
Meshes don't work yet.


shaun-edwards:
@TheDash, do you intend on fixing the meshes for this PR?

UltronDestroyer:
Not for this PR, but I will make another one that fixes the meshes. I'm working on a project where they are necessary, and so it should happen sometime this week.

shaun-edwards:
I'm not sure the visualization package is where the gazebo launch files should go?  We do not have a ROS-I standard for simulation packages, but the UR is an example, see [here](https://github.com/ros-industrial/universal_robot/tree/hydro-devel/ur_gazebo).  Perhaps we should create a `robotiq_s_model_gazebo` package for the launch files in this PR.  Thoughts?

UltronDestroyer:
Hm, strange that the visualization contains the URDF and such where most robots contain their urdf' inside a _description package. If we move this to a gazebo package I would suggest moving the _visualization to a _description package to follow standard formatting

shaun-edwards:
ROS-Industrial has a standard for package layout, see [here](http://wiki.ros.org/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories).  It's slightly different from traditional ROS packages in order to support various models and variations available from the different vendors.  The Robotiq repo pre-dates this standard, which is why it doesn't conform.  When bringing other packages into conformance, we have tried to maintain backwards compatibility.  To do this, we typically make copies of package files and then mark the existing package as [deprecated](http://www.ros.org/reps/rep-0127.html#deprecated).  This allows users one or two ROS versions to adapt to the changes.

UltronDestroyer:
Ah, so the issue with meshes is:

this package uses .stl files. Gazebo renders .stl as a solidly coloured link (which is why it appears all white). .dae files are required to have a nicely decorated model. Is there any way to get those?

gavanderhoorn:
Does gazebo not support the urdf `material` elements? Collada files are nicer, but also more work to create. Adding a `color` element to the links would add colour to the meshes, without requiring new ones.

UltronDestroyer:
Those are already in the URDF file. I think the .stl files are overriding it though since they're set for the mesh. I'll see what happens when I remove the .stl meshes for visualization

Edit: It looks like I can't remove the .stl, since they are used for geometry and collisions. The <material> color tag is set, but isn't being visualized.

The way the PR2 does it is use .dae files for the geometry tags, and .stl for the collision tags.

UltronDestroyer:
.

gavanderhoorn:
According to the tutorial page [Using a URDF in Gazebo](http://gazebosim.org/tutorials/?tut=ros_urdf), section "Links", `material/color` tags should work.

UltronDestroyer:
Do you mean this?

 Unfortunately, this method of specifying link colors does not work in Gazebo as it adopts OGRE's material scripts for coloring and texturing links. Instead, a Gazebo material tag must be specified for each link, such as:

```
  <gazebo reference="link1">
    <material>Gazebo/Orange</material>
  </gazebo>
```

gavanderhoorn:
Yes, that is probably how it should be done. Or at least, that's what I've seen a lot.

UltronDestroyer:
Ok, that seemed to give me color. It won't look exactly how it would with a .dae file, but at least its better than white.

UltronDestroyer:
I've added as much as I can. If we want to improve the look further of the Gazebo model then I suggest we use .dae files

gavanderhoorn:
I'm not sure, but wouldn't we want the Gazebo compatibility _in addition_ to the colours working in RViz and other tools that do support URDF `material` and `color` elements?

robotiq_s_model_visualization/cfg/s-model_finger_articulated_macro.xacro
```diff
@@ -34,20 +28,22 @@ finger(i.e. finger_1, etc...).
                 <inertia ixx="0.01" ixy="-0.00002" ixz="0.00001" iyy="0.0008" iyz="0" izz="0.001" />
             </inertial>         
 		</link>
+  <gazebo reference="${prefix}link_0">
+                            <material>Gazebo/Black</material>
+                        </gazebo>
+
```
gavanderhoorn: indent?

robotiq_s_model_visualization/cfg/s-model_finger_articulated_macro.xacro
```diff
@@ -55,6 +51,9 @@ finger(i.e. finger_1, etc...).
                 <inertia ixx="0.001859" ixy="-0.000376" ixz="0.000028" iyy="0.012756" iyz="0" izz="0.0024" />
             </inertial>
 		</link>
+ <gazebo reference="${prefix}link_1">
+                            <material>Gazebo/Black</material>
+                        </gazebo>
```
gavanderhoorn: same.

robotiq_s_model_visualization/cfg/s-model_finger_articulated_macro.xacro
```diff
@@ -81,20 +78,21 @@ finger(i.e. finger_1, etc...).
                 <inertia ixx="0.001859" ixy="-0.000376" ixz="0.000028" iyy="0.012756" iyz="0" izz="0.0024" />
             </inertial>
 		</link>
+ <gazebo reference="${prefix}link_2">
+				<material>Gazebo/Black</material>
+                                </gazebo>
```
gavanderhoorn: ..

UltronDestroyer:
Ah, so leave those elements in?

gavanderhoorn:
@TheDash wrote:

> Ah, so leave those elements in?

I'd say yes. RViz will ignore the Gazebo defs, Gazebo ignores the URDF defs (iiuc).

UltronDestroyer:
https://github.com/ros-industrial/robotiq/blob/groovy-devel/robotiq_s_model_visualization/cfg/s-model_articulated_macro.xacro

I've fixed a bug where the <inertial> tag is outside of the <link></link> elements of the robotiq_palm. This causes Gazebo to not load a controller that implements any joint which combines robotiq_palm since the robotiq_palm does not have physical properties. See http://answers.ros.org/question/102276/this-robot-has-a-joint-named-foo-which-is-not-in-the-gazebo-model/

And check the latest commit in my branch

UltronDestroyer:
FYI this too 
http://answers.gazebosim.org/question/6944/urdf-joints-not-appearing-in-gazebo/

answered Sep 3 '14
nkoenig gravatar image nkoenig
5074 ● 5 ● 21 ● 44
http://gazebosim.org/

All the the joints shown in your snippet are fixed. Gazebo merges all links connected by fixed joints into a single link. In your case, no joints will be present.

So that means the palm joint (which is fixd right now) must be revolute to work in Gazebo. I don't know why it's fixed anyway, in the real model it is revolute no?

gavanderhoorn:
Right. So if your satisfied with the changes you've made, could you perhaps squash them into one (or maybe two) commits? Thanks.

UltronDestroyer:
I've added some more things (transmission files, renaming of joint type).
I'll make a squash later on today. But the overall changes will allow the
robotiq gripper to be ros_control enabled and work in gazebo.

gavanderhoorn:
?

UltronDestroyer:
Sorry. I remade the package. I'll PR in a moment.

PR #16, opened 2015-01-09
Robotiq C Model Gripper Visualization

Jmeyer1292:
Adapted the Robotiq c-model gripper design to ROS-Industrial.


gavanderhoorn:
@Jmeyer1292: you could easily squash your three commits into one, seeing as they are only 'fixup commits' of the first one?

Jmeyer1292:
I looked up the squash procedure and I believe I have correctly combined my three commits into one. Thanks for you patience with me: I'm excited contribute but obviously learning good etiquette. Please let me know if there's anything else I can do.

gavanderhoorn:
@Jmeyer1292: very nice, thank you. This keeps the history of the repository a bit cleaner.

For the future: I normally also rewrite the commit msg to be one single description of what the changes are, but that is ok for now.

robotiq_c_model_visualization/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<package>
+  <name>robotiq_c_model_visualization</name>
+  <version>1.0.0</version>
+  <description>The robotiq_c_model_visualization package</description>
+
+  <maintainer email="jonathan.meyer@swri.org">Jonathan Meyer</maintainer>
+  <author>Kel Guerin (Robotiq Inc.)</author>
```
shaun-edwards: This isn't correct.  Kel Guerin does not work at Robotiq.

robotiq_c_model_visualization/package.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0"?>
+<package>
+  <name>robotiq_c_model_visualization</name>
+  <version>1.0.0</version>
```
shaun-edwards: Does this version match the versions of all other robotiq packages?  It may not matter since the `robotiq` repo has never been release.  It will make this easier though.

robotiq_c_model_visualization/package.xml
```diff
@@ -0,0 +1,20 @@
...
+
+  <license>BSD</license>
+
+  <url type="website">http://ros.org/wiki/robotiq</url>
+
+  <buildtool_depend>catkin</buildtool_depend>
+  <build_depend>urdf</build_depend>
+  <run_depend>urdf</run_depend>
+
+  <export>
```
shaun-edwards: export tags can be deleted.

shaun-edwards:
@Jmeyer1292, can you paste a screenshot into a comment under the PR?

robotiq_c_model_visualization/package.xml
Jmeyer1292: Fixed. I apologize, I must have got cross eyed looking at the other packages.

robotiq_c_model_visualization/package.xml
Jmeyer1292: As far as I can tell, all of the robotiq packages are labeled version 1.0.0.

robotiq_c_model_visualization/package.xml
Jmeyer1292: Done

Jmeyer1292:
Is this what you meant?
robotiq_c_model

Jmeyer1292:
Also, would you prefer me to squash the fixed commits again?

gavanderhoorn:
@Jmeyer1292: personally I'd say yes, as the second commit is again a fix-up.

UltronDestroyer:
Thanks for this @Jmeyer1292

shaun-edwards:
+1 from me!  Others?

gavanderhoorn:
- would perhaps be nice to include some launch files to be able to launch a stand-alone rviz instance with just the model
- why is there a build & run depend on `urdf`, while this package does not contain any compilable artefacts. Wouldn't a run_depend on `xacro` make more sense (but then only if there are launch files that actually load the xacro, using `xacro`)?
- the xacro specifies a value of 1000 for the `effort` limit. Is this a real value?
- package description in manifest is non-informative

nicolaslauzier:
Hi guys,

Sorry for stepping up late in the discussion, I am not as active as I should be here! This seems to me like a very nice add-on that will be useful for the gripper's users! 

The only point that I would add is that we recently launched a new version of the 2F-85 gripper (http://robotiq.com/products/industrial-robot-gripper/) which is similarly controlled (although it is now possible to control it with a serial/USB connection without external controller) and which has improved specs (twice the force while being lighter). Anyway, the fingers are similar but the casing is quite different, which prompts me to ask if it is possible to change the name of the visualization package from "C" to "C2", to distinguish it from the newly released version 3 of the gripper (which would be "C3" if the package was remade for this version).

Also, "C" and "S" are the old names for the "2F-85" and "3F" grippers but this doesn't need to be changed in the short term as long as it doesn't add confusion.

Again, thanks for this very nice contribution!

Jmeyer1292:
@gavanderhoorn I added a visualization launch file and updated the package file. The effort limits were 1000.0 in the https://github.com/jhu-lcsr-forks/robotiq repo that I adapted this from. This mechanical specification of 2F-85 states that the max grip force is 200N but I'm not entirely sure how that will propogate back through the joints of the fingers.

@nicolaslauzier Thank you for the compliment, but I'm just adapting these files. There's more stuff coming down the pipe soon for the ROSI Robotiq package so I think it would be a good idea to get the naming conventions ironed out. I'm brand new here, though, so what does everyone else think?

gavanderhoorn:
@Jmeyer1292 wrote:

>  I added a visualization launch file [..]

Would it perhaps be an idea to reuse the `test_....launch` files we use in other ROS-Industrial packages? Functionally equivalent to what you added, but the naming is semi-standardised across packages. See [abb_irb2400_support/launch/test_irb2400.launch](https://github.com/ros-industrial/abb/blob/hydro-devel/abb_irb2400_support/launch/test_irb2400.launch) fi.

> [..] and updated the package file.

You forgot to update `CMakeLists.txt`.

As to the renaming: we try to keep things backward compatible. You never know in what way people are using the packages that are there now. Renaming can be done, but should be 'additive', and follow a _tick-tock_ approach.

Introduce a new package (essentially a duplicate), that follows your new naming scheme, but is identical in content and features otherwise. Keep it next to the old one, which we mark _deprecated_. Then in the next distribution, you can remove the old package. The deprecated package should also provide a hint as to which package could be used as a replacement. As an example, see `package.xml` in [abb_common](https://github.com/ros-industrial/abb/tree/hydro-devel/abb_common) or [fanuc_assets](https://github.com/ros-industrial/fanuc/tree/hydro-devel/fanuc_assets).

Jmeyer1292:
@gavanderhoorn I updated the CMakeLists and renamed the launch file to test_robotiq_c_model.launch. The re-naming shouldn't be an issue here though because this is a new package to this repository right?

gavanderhoorn:
>  I updated the CMakeLists and renamed the launch file to test_robotiq_c_model.launch.

Great, now make the display of the `joint_state_publisher` gui `true` by default :).

> The re-naming shouldn't be an issue here though because this is a new package to this repository right?

No, you're right. I just thought I'd explain the general procedure regarding renaming ROS-Industrial packages.

projet-mask:
@Jmeyer1292 Please note that the 200N specification refers to the new gripper version (C3). The C2 version had 100N as a maximum force.

nicolaslauzier:
@Jmeyer1292 Please note that the 200N spec refers to the new version of the gripper (C3). The C2 gripper's maximum force was 100N.

Jmeyer1292:
@nicolaslauzier Thanks for pointing that out. I haven't changed efforts yet, but I did rename the package to robotiq_c2_model_visualization.

@gavanderhoorn I set it to true by default.

gavanderhoorn:
I just noticed that the package doesn't seem to include collision meshes. Is that on purpose (perhaps because of more detailed collision checks)?

Jmeyer1292:
@gavanderhoorn They were not present in the repo I adapted them from: https://github.com/jhu-lcsr-forks/robotiq. A quick search for other repos and Google doesn't seem to turn them up either.

gavanderhoorn:
I'm not sure how detailed the current meshes are. If they have low enough faces/vertices count, then adding additional collision meshes might not be necessary.

Jmeyer1292:
@gavanderhoorn I'm not sure how to judge. I tried to follow through on the tutorial at http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot that suggests using MeshLab to simplify STL files. Meshlab 1.3 seems to immediately crash when I load these files though.

gavanderhoorn:
Blender can open them. My Meshlab crashes as well.

`robotiq_85_gripper_base.dae` has 7k vertices and 14k faces. A convex hull of the same mesh has 203 and 402 respectively. Not sure about the rest, but that seems like a significant enough reduction to me.

Workflow used: Open `.dae` in Blender, export as `.stl`, use Meshlab to remesh as convex hull (take care to not introduce inconsistent normals). Save as `.stl` again.

Perhaps Blender can create convex hulls, I don't know (left as exercise for the reader).

gavanderhoorn:
I'm not sure whether we want to do this for all meshes btw: I can imagine that fi for the fingers, we want precise collision checks. Not sure there what the best approach is.

Jmeyer1292:
@gavanderhoorn I will follow your procedure and make some simplified meshes. If users want detailed models, they can use the visual meshes. 

A note for the future to anyone who is curious: the blender debian in the Ubuntu 12.04 packages was not compiled with Collada extensions.

gavanderhoorn:
Just to be sure: include both the meshes you have now _and_ a set of 'collision quality' meshes. The latter are for collision avoidance/detection algorithms (such as FCL in MoveIt), the former for visualisation purposes (and can be as detailed as you like).

See other ros-i support pkgs for a suggested layout of the `meshes` directory (or [this](http://wiki.ros.org/Industrial/Tutorials/SuggestedPackageLayoutNewRepositories) unfinished tutorial (ignore the wrong model vs type distinction)).

Jmeyer1292:
@gavanderhoorn It's done. I somehow managed to turn the models 90 degrees about Z in the process, but putting a corresponding counter rotation into the URDF seems to make it behave appropriately.

gavanderhoorn:
Just did a quick test. Meshes look ok. If we are all happy with naming, lets merge this as it is right now?

shaun-edwards:
@Jmeyer1292....I recommend we remove mass/inertia values.  Very few ROS-I packages have these values because its hard to verify them.

Jmeyer1292:
@shaun-edwards  @asisbot What about taking the published values and having one inertia tag on the base piece that specifies the published COM and Inertial matrix? Is that a valid solution?

shaun-edwards:
These values are used in Gazebo, but I'm not sure if what you propose would cause any issues.  I'm sure these values are used in other ways as well, so it's hard to predict what effect your proposal would have.  I'd prefer to stick with individual link COM and inertia.  We could ask robotiq to provide these values.  We could accept this PR (with COM and inertia removed) and log an issue against @nicolaslauzier

Jmeyer1292:
@shaun-edwards I removed the intertial tags

shaun-edwards:
+1.... @nicolaslauzier, please give the final approval.

gavanderhoorn:
If I may suggest squashing all fixup commits again?

nicolaslauzier:
Everything looks fine to me. @gavanderhoorn , should I wait before merging the pull request?

I'll log internally that we should publish the inertial data of the 2F gripper. I cannot give a specific schedule for this though.

gavanderhoorn:
@nicolaslauzier: personally I like to avoid merging many "fixed this-and-this after PR review" commits, which is why I suggested squashing the 'fixups' into the 'initial commit of robotiq_..' commit.

It's up to you really. I think @Jmeyer1292 returns to work in about 2hrs, and squashing shouldn't take too long, so if you can wait .. :).

Jmeyer1292:
@asisbot I noticed you forked my fork of robotiq, so out of politeness I'd like to make sure you're cool with a rebase.

Jmeyer1292:
@gavanderhoorn @shaun-edwards @nicolaslauzier I've squashed all the commits into one.

gavanderhoorn:
Not trying to be _that guy_, but it seems the [urdf](https://github.com/ros-industrial/robotiq/pull/16/files#diff-24) hasn't been updated. If you do that I think we could merge this?

Jmeyer1292:
@gavanderhoorn I'm sorry, I overlooked that. I'll fix it shortly.

Jmeyer1292:
@gavanderhoorn The URDFs are updated with the latest xacro files and the fix is squashed.

shaun-edwards:
I'm glad @gavanderhoorn is _that guy_... @Jmeyer1292, is it good to go now?

Jmeyer1292:
@shaun-edwards I believe so.

PR #20, opened 2015-02-09
Robotiq Action Server

Jmeyer1292:
This request adds a simple_action_server which communicates with a named gripper using the control_msgs::GripperCommandAction type. 

The exact specification of your gripper can be defined in code or in the launch file. This includes min/max values for both gripper position and gripper effort. The robotiq accepts position requests in the form of an unsigned byte 0-255, however it closes prior to 255. After the fingers close, they push forward slightly. I chose to define min_position as having a negative number that corresponds to the gap size decreasing linearly past zero. If you don't like this, you can set your min gap to zero and the program will linearly interpolate the resulting command value accordingly.

The result object will report stalled if the gripper was not able to reach its position for any reason, including object detection. The GripperCommandAction interface doesn't allow for the more fine grained robotiq information.

Thanks for looking at this!

### DIFF
--- a/robotiq_action_server/src/robotiq_c_model_action_server.cpp
+++ b/robotiq_action_server/src/robotiq_c_model_action_server.cpp
@@ -1,0 +1,192 @@
+/**
+ * ActionServer interface to the control_msgs/GripperCommand action
+ * for a Robotiq C-Model device
+ */
+
+#include "robotiq_action_server/robotiq_c_model_action_server.h"
+
+// To keep the fully qualified names managable
+namespace ras = robotiq_action_server;
+
+//Anonymous namespaces are file local -> sort of like global static objects
+namespace
+{
+  using namespace ras;
+
+  /*  This struct is declared for the sole purpose of being used as an exception internally
+      to keep the code clean (i.e. no output params). It is caught by the action_server and 
+      should not propogate outwards. If you use these functions yourself, beware.
+  */
+  struct BadArgumentsError {};
+
+
+  GripperOutput goalToRegisterState(const GripperCommandGoal& goal, const CModelGripperParams& params)
+  {
+    GripperOutput result;
+    result.rACT = 0x1; // active gripper
+    result.rGTO = 0x1; // go to position
+    result.rATR = 0x0; // No emergency release
+    result.rSP = 128; // Middle ground speed
+    
+    if (goal.command.position > params.max_gap_ || goal.command.position < params.min_gap_)
+    {
+      ROS_WARN("Goal gripper gap size is out of range(%f to %f): %f m",
+               params.min_gap_, params.max_gap_, goal.command.position);
+      throw BadArgumentsError();
+    }
+    
+    if (goal.command.max_effort < params.min_effort_ || goal.command.max_effort > params.max_effort_)
+    {
+      ROS_WARN("Goal gripper effort out of range (%f to %f N): %f N",
+               params.min_effort_, params.max_effort_, goal.command.max_effort);
+      throw BadArgumentsError();
+    }
+
+    double dist_per_tick = (params.max_gap_ - params.min_gap_) / 255;
+    double eff_per_tick = (params.max_effort_ - params.min_effort_) / 255;
+
+    result.rPR = static_cast<uint8_t>((goal.command.position - params.min_gap_) / dist_per_tick);
+    result.rFR = static_cast<uint8_t>((goal.command.max_effort - params.min_effort_) / eff_per_tick);
+
+    return result;
+  }
+
+  /*  This function is templatized because both GripperCommandResult and GripperCommandFeedback consist
+      of the same fields yet have different types. Templates here act as a "duck typing" mechanism to avoid
+      code duplication.
+  */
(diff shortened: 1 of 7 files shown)


UltronDestroyer:
Thanks Johnathan for your work here. I was excited when I got the email about it. I have a question though, how is it different from http://wiki.ros.org/gripper_action_controller?

Jmeyer1292:
In all honesty, I was not aware of the implementation you pointed out. They seem to be trying to create a generic interface to all grippers and they might be keeping state in the action server.

From the end user perspective there shouldn't be much of a difference. I do make the assumption that your running a control node that accepts input/output commands on the <name>/input and <name>/output topics. 

I'll dig into their code when I have time. Are you using this interface already? I can support you much faster with my own implementation because I wrote it and its a lot simpler, but it might be a good idea to adapt my Hardware interface layer to their code.

UltronDestroyer:
I'm using that interface for MoveIt! of the Robotiq 3 finger gripper, but it's limited. It can only open and close relative to a defined planar point. I would like to have an action server that is able to control 3 fingers, ideally. And a server that hopefully supports underactuation.

robotiq_action_server/src/robotiq_c_model_action_server.cpp
```diff
@@ -0,0 +1,194 @@
...
+ras::CModelGripperActionServer::CModelGripperActionServer(const std::string& name, const CModelGripperParams& params)
+  : nh_()
+  , as_(nh_, name, false)
+  , action_name_(name)
+  , gripper_params_(params)
+{
+  as_.registerGoalCallback(boost::bind(&CModelGripperActionServer::goalCB, this));
+  as_.registerPreemptCallback(boost::bind(&CModelGripperActionServer::preemptCB, this));
+
+  state_sub_ = nh_.subscribe(name + "/input", 1, &CModelGripperActionServer::analysisCB, this);
```
shaun-edwards: Why use the `name` instead of simply remapping arguments?

robotiq_action_server/src/robotiq_c_model_action_server.cpp
```diff
@@ -0,0 +1,194 @@
+/**
+ * ActionServer interface to the control_msgs/GripperCommand action
+ * for a Robotiq C-Model device
+ */
+
+#include "robotiq_action_server/robotiq_c_model_action_server.h"
+
+// To keep the fully qualified names managable
+namespace ras = robotiq_action_server;
```
shaun-edwards: This namespace aliasing should not be required.  The only namespace in any file should be the package namespace.

shaun-edwards:
@Jmeyer1292, Do you know if this action server used by MoveIt directly?

Jmeyer1292:
It's the action interface used by moveit's simple_controller_manager (see https://github.com/ros-planning/moveit_plugins), if that's what you mean. How exactly to interface it in MoveIt, I'm not sure yet.

robotiq_action_server/src/robotiq_c_model_action_server.cpp
Jmeyer1292: I'll change to the following format: 

namespace <ns>
{
}

robotiq_action_server/src/robotiq_c_model_action_server.cpp
Jmeyer1292: My thought here was that maybe someone would want to create multiple instances of the action_server inside the same process. Would the preferable method be to set the topic names to "input" and "output" and let users remap?

shaun-edwards: Creating multiple instances within a single node is not typical.  It also means that remapping cannot be done easily.  It would require intimate knowledge about how names are concatenated.  Instead of creating multiple instances in a single process, launching multiple nodes is preferred.

robotiq_action_server/src/robotiq_c_model_action_server_node.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+namespace
+{
+  // Defines a default for the c2 model 85 gripper
+  robotiq_action_server::CModelGripperParams c2_85_defaults()
+  {
+    robotiq_action_server::CModelGripperParams params;
+    params.min_gap_ = -.017;
+    params.max_gap_ = 0.085;
+    params.min_effort_ = 30.0; // This is a guess. Could not find data with quick search.
```
fsuarez6: min_effort: ~40 Nm. Value obtained using a ATI Gamma F/T Sensor.

shaun-edwards:
@Jmeyer1292 ping?

Jmeyer1292:
@shaun-edwards Like the other PR, I think this one is good to go if it's useful. Also like the other PR, I haven't touched it in forever so it'd probably be best to find the hardware and try to test.

Jmeyer1292:
Though the current server only works for the two finger jobs, it appears to work as desired. I think it's good to go.

shaun-edwards:
+1.... @nicolaslauzier, this PR is in support of higher level/asynchronous control of robotiq grippers.

PR #21, opened 2015-02-09
Ethercat interface

Jmeyer1292:
This pull request adds EtherCAT protocol support for the Robotiq gripper. It defines an EtherCatManager class which handles communication with the ethercat network via the SOEM library which this depends on. Adding this library will be handled soon, but can currently be found in my fork https://github.com/Jmeyer1292/ethercat-soem. 

Each device defines an ethercat client that communicates directly with the manager, the latter of which only deals in bytes. 

**\* The main limitation of this implementation is that all devices on a given ethercat network need to be handled in a single process. ***

On top of this system is a c model gripper control node made in the style of the existing python nodes. See the launch file added to robotiq_c_model_control for more details.

Despite its limitations, I think this will be effective for many users. My Robotiq EtherCAT control box only has a single ethernet connection so it will only work by itself anyway. If additional capability is required, I have ideas for expanding capabilities but I wanted to get this out to folks.

Please note that:
1. VMs probably won't work
2. You need a dedicated ethernet card for the EtherCAT network. Don't use eth0.
3. You'll probably need to run "sudo setcap cap_net_raw+ep <filename>" on the c_model_ethercat_node before it will work. Otherwise you'll get a "Could not initialize SOEM" error.

### DIFF
--- a/robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_client.cpp
+++ b/robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_client.cpp
@@ -5,9 +5,10 @@
 // See Robotiq's documentation for the register mapping
 
 // An effort to keep the lines less than 100 char long
-namespace c_model = robotiq_c_model_control;
+namespace robotiq_c_model_control
+{
 
-c_model::CModelEtherCatClient::CModelEtherCatClient(robotiq_ethercat::EtherCatManager& manager, 
+CModelEtherCatClient::CModelEtherCatClient(robotiq_ethercat::EtherCatManager& manager, 
                                                     int slave_no)
   : manager_(manager)
   , slave_no_(slave_no)
@@ -16,7 +17,7 @@ c_model::CModelEtherCatClient::CModelEtherCatClient(robotiq_ethercat::EtherCatMa
 /*
   See support.robotiq.com -> manual for the register output meanings
 */
-void c_model::CModelEtherCatClient::writeOutputs(const GripperOutput& output)
+void CModelEtherCatClient::writeOutputs(const GripperOutput& output)
 {
   uint8_t map[6] = {0}; // array containing all 6 output registers
 
@@ -33,7 +34,7 @@ void c_model::CModelEtherCatClient::writeOutputs(const GripperOutput& output)
   }
 }
 
-c_model::CModelEtherCatClient::GripperInput c_model::CModelEtherCatClient::readInputs() const
+CModelEtherCatClient::GripperInput CModelEtherCatClient::readInputs() const
 {
   uint8_t map[6];
 
@@ -57,7 +58,7 @@ c_model::CModelEtherCatClient::GripperInput c_model::CModelEtherCatClient::readI
   return input;
 }
 
-c_model::CModelEtherCatClient::GripperOutput c_model::CModelEtherCatClient::readOutputs() const
+CModelEtherCatClient::GripperOutput CModelEtherCatClient::readOutputs() const
 {
   uint8_t map[6];
   for (unsigned i = 0; i < 6; ++i)
@@ -75,3 +76,4 @@ c_model::CModelEtherCatClient::GripperOutput c_model::CModelEtherCatClient::read
 
   return output;
 }
+} // end of robotiq_c_model_control namespace
--- a/robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_client.cpp
+++ b/robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_client.cpp
@@ -1,0 +1,77 @@
+#include "robotiq_c_model_control/c_model_ethercat_client.h"
+
+#include "robotiq_ethercat/ethercat_manager.h"
+
+// See Robotiq's documentation for the register mapping
+
+// An effort to keep the lines less than 100 char long
+namespace c_model = robotiq_c_model_control;
+
+c_model::CModelEtherCatClient::CModelEtherCatClient(robotiq_ethercat::EtherCatManager& manager, 
+                                                    int slave_no)
+  : manager_(manager)
+  , slave_no_(slave_no)
+{}
+
+/*
+  See support.robotiq.com -> manual for the register output meanings
+*/
+void c_model::CModelEtherCatClient::writeOutputs(const GripperOutput& output)
+{
+  uint8_t map[6] = {0}; // array containing all 6 output registers
+
+  // Pack the Action Request register byte
+  map[0] = (output.rACT & 0x1) | ((output.rGTO << 0x3) & 0x8) | ((output.rATR << 0x4) & 0x10);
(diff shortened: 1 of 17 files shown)


shaun-edwards:
@Jmeyer1292, we should transfer SOEM to ROS-Industrial before this PR is accepted.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_client.cpp
```diff
@@ -0,0 +1,77 @@
+#include "robotiq_c_model_control/c_model_ethercat_client.h"
+
+#include "robotiq_ethercat/ethercat_manager.h"
+
+// See Robotiq's documentation for the register mapping
+
+// An effort to keep the lines less than 100 char long
+namespace c_model = robotiq_c_model_control;
```
shaun-edwards: Can you can create a namespace section, instead of aliasing the namespace.

```
namespace robotiq_c_model_control
{
<code>
}
```

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
```diff
@@ -0,0 +1,81 @@
...
+  {
+    // Check to see if resetting is required? Or always reset?
+    GripperOutput out;
+    out.rACT = 0x1;
+    client.writeOutputs(out);
+  }
+
+  // Sorry for the indentation, trying to keep it under 100 chars
+  ros::Subscriber sub = 
+        nh.subscribe<GripperOutput>("/CModelRobotOutput", 1,
```
shaun-edwards: topic names should not be globally namespaced.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
```diff
@@ -0,0 +1,81 @@
...
+                                    boost::bind(changeCallback, boost::ref(client), _1));
+
+  ros::Publisher pub = nh.advertise<GripperInput>("/CModelRobotInput", 100);
```
shaun-edwards: Global namespace.

robotiq_ethercat/src/ethercat_manager.cpp
```diff
@@ -0,0 +1,226 @@
...
+      }
+
+      usleep(THREAD_SLEEP_TIME);
+    }
+  }
+
+} // end of anonymous namespace
+
+
+robotiq_ethercat::EtherCatManager::EtherCatManager(const std::string& ifname)
```
shaun-edwards, line 111: Could be `robotiq_ethercat` namespace.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+int main(int argc, char** argv)
+{
+  using robotiq_ethercat::EtherCatManager;
+  using robotiq_c_model_control::CModelEtherCatClient;
+
+  typedef CModelEtherCatClient::GripperOutput GripperOutput;
+  typedef CModelEtherCatClient::GripperInput GripperInput;
+
+  ros::init(argc, argv, "robotiqCModel");
```
shaun-edwards: Node name should be `lower_case`

robotiq_ethercat/src/ethercat_manager.cpp
Jmeyer1292, line 111: Anonymous name-spacing keep functions and variables private inside the compilation unit. It's equivalent to declaring everything static. I didn't want to expose those functions.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
Jmeyer1292: These topic names were chosen to match the names in the equivalent python nodes. If I change them, then we'll need to create launch files or modify those scripts to make everything work together.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
Jmeyer1292: Again, I was matching the existing nodes. I will change if that is preferred.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_client.cpp
Jmeyer1292: I prefer to explicitly qualify all the names in source files, but I admit its really ugly to have to remap namespaces. I'll change this.

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
shaun-edwards: Globally namespaced topics are a bug.  We can fix it here and then log an issue for the python scripts.

robotiq_ethercat/src/ethercat_manager.cpp
shaun-edwards, line 111: OK

robotiq_c_model_control/src/robotiq_c_model_control/c_model_ethercat_node.cpp
shaun-edwards: The node name is rarely used.  Launch files typically rename them.  It's safe to change it and good to adhere to naming conventions.

Jmeyer1292:
I'm fixing these issues, but please give me a chance to test before moving forward.

gavanderhoorn:
Just a general, high-level comment: would it make sense to make use of a pkg like [ros_ethercat](https://github.com/shadow-robot/ros_ethercat) for this?

shaun-edwards:
ROS Ethercat may be a better solution.  The current solution is based on some existing development.  I'd like to suggest we move ahead with this PR since something is better than nothing.  We can add an issue to investigate the `ros_ethercat` library.

shaun-edwards:
@Jmeyer1292 ping?

Jmeyer1292:
@shaun-edwards I think this is good to go, but I haven't messed with it since early Feb. I'll see if I can't hunt down the hardware for this and put it back together to test it.

Jmeyer1292:
I've tested it and it seems to work just fine. I added a readme file to cover the common issues mentioned in the header of this PR.

Jmeyer1292:
I'm happy to update this pull request with the contents of one made against by feature branch by Mr. Wilson Ko who has added S model gripper control to the Robotiq ethercat driver set. I have tested it and found it to work nicely.

@shaun-edwards ping

shaun-edwards:
+1 from me... @nicolaslauzier, this PR enabled EtherCAT support.  I assume you are in support of merging this.

PR #25, opened 2015-05-13
ADD: WrenchStamped publisher, ROS_INFO to see if node is running

lorenz-halt:
... and node can be stopped with CTRL-C


shaun-edwards:
In addition to other minor fixes, does this simply publish a standard message (Wrench) with data already contained within the custom robotiq message?

lorenz-halt:
Exactly

shaun-edwards:
+1 from me... @nicolaslauzier, what do you think

nicolaslauzier:
Yes, I think we should have output a Wrench message from the beginning. This was not done simply because we didn't know this standard message. Also, thanks for the ros::ok() addition.

Issue #26, opened 2015-05-14
catkin_lint errors


gavanderhoorn:
```
robotiq_c_model_control: CMakeLists.txt(8): error: variable EXECUTABLE_OUTPUT_PATH is modified
robotiq_c_model_control: CMakeLists.txt(10): error: variable LIBRARY_OUTPUT_PATH is modified
robotiq_c_model_control: notice: package description starts with boilerplate 'Package to'
robotiq_force_torque_sensor: error: message dependency 'std_msgs' is not listed in catkin_package()
robotiq_force_torque_sensor: warning: package 'std_msgs' should be listed in catkin_package()
robotiq_force_torque_sensor: notice: package description starts with boilerplate 'Package for'
robotiq_force_torque_sensor: notice: target 'rq_sensor' is not installed
robotiq_modbus_rtu: CMakeLists.txt(7): error: variable EXECUTABLE_OUTPUT_PATH is modified
robotiq_modbus_rtu: CMakeLists.txt(9): error: variable LIBRARY_OUTPUT_PATH is modified
robotiq_modbus_tcp: CMakeLists.txt(7): error: variable EXECUTABLE_OUTPUT_PATH is modified
robotiq_modbus_tcp: CMakeLists.txt(9): error: variable LIBRARY_OUTPUT_PATH is modified
robotiq_s_model_control: error: unconfigured build_depend on 'message_generation'
robotiq_s_model_control: error: unconfigured build_depend on 'rospy'
robotiq_s_model_control: error: unconfigured build_depend on 'message_generation'
robotiq_s_model_control: CMakeLists.txt(4): error: missing COMPONENTS keyword before 'rospy'
robotiq_s_model_control: CMakeLists.txt(4): error: missing COMPONENTS keyword before 'message_generation'
robotiq_s_model_control: CMakeLists.txt(7): error: variable EXECUTABLE_OUTPUT_PATH is modified
robotiq_s_model_control: CMakeLists.txt(9): error: variable LIBRARY_OUTPUT_PATH is modified
robotiq_s_model_control: notice: package description starts with boilerplate 'Package to'
robotiq_s_model_visualization: error: unconfigured build_depend on 'urdf'
robotiq_s_model_visualization: CMakeLists.txt(4): error: missing COMPONENTS keyword before 'urdf'
```

gavanderhoorn:
Update (`catkin_lint -W2 robotiq` run against repo at f3e1c987):

```
[...]

PR #27, opened 2015-05-14
Fix for issues 10 & 22: remove wrong and unneeded 'requires' entry from setup args

gavanderhoorn:
This should make the docjob(s) run to completion again and make `catkin_make install` work again.

I opted to remove the entire `requires` entry as it is unneeded:
- catkin pkgs should be installed using catkin, not `setup.py`
- dependency is already expressed in the pkg manifest

Upon acceptance this should probably be cherry-picked into `groovy-devel` and `hydro-devel`.

Perhaps a quick re-release of `hydro-devel` is then in order, as the buildfarm will be disabled soon.

### DIFF
--- a/robotiq_modbus_rtu/package.xml
+++ b/robotiq_modbus_rtu/package.xml
@@ -5,20 +5,15 @@
   <maintainer email="sedwards@swri.org">Shaun Edwards</maintainer>
   <maintainer email="kphawkins@gatech.edu">Kelsey Hawkins</maintainer>
   <license>BSD</license>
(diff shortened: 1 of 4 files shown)


gavanderhoorn:
Btw: the manifests and build scripts of all pkgs could use some work. See also #26.

gavanderhoorn:
Ping?

The Hydro buildfarm is nearing EOL, would be nice if the docs could be generated at least once?

gavanderhoorn:
I'd recommend this be cherry picked into the `groovy-devel` and `hydro-devel` branches as well.

Or would you prefer I open PRs against those branches?

PR #28, opened 2015-05-21
PR27 but for groovy-devel

gavanderhoorn:
.


nicolaslauzier:
Thanks for these modifications!

PR #35, opened 2015-07-06
Syncing changes from hydro to indigo branch

shaun-edwards:
Seems the hydro/indigo branch are out of sync.  This pulls several changes from hydro into indigo (the main development branch).


shaun-edwards:
@gavanderhoorn, it looks like you had some comments above that got automatically added to the PR.  Not sure if these were addressed.  Maybe they should be before we do this PR to sync things back up.

gavanderhoorn:
Not sure what is going on there. Perhaps I commented on the commit itself, instead of only within the context of the PR.

Afaict, the pkg manifests are ok.

gavanderhoorn:
Ah, now I remember: in that particular commit, a duplicate `run_depend` was introduced. Later commits (some of my own actually) have rectified that.

I've removed the two comments (from the commits themselves).

shaun-edwards:
@nicolaslauzier, can you please merge this?  BTW, this mis-sync was caused by myself.  @Jmeyer1292 PR was issued against hydro and I missed it.

PR #36, opened 2015-07-06
Added code to fix rq_sensor hang on ctrl-c

futureneer:
This fixes rq_sensor's hang when you control-c after rosrun-ing it from the command line. This also adds some verbosity so it is clear what the sensor is doing as it boots up. These might make more sense as loginfo rather than logwarn.

Cheers,
Kel


shaun-edwards:
@futureneer, can you please submit this to the indigo-devel branch?  Some of these changes have already been done on that branch.

futureneer:
Ok, will do.

Issue #37, opened 2015-07-06
Improper clean up of serial interface on ctrl-c of rq_sensor


futureneer:
Hello,
When ctrl-c is invoked on rq_sensor, I believe the serial interface is not being cleaned up properly. I think this because when the sensor is restarted shortly after the ctrl-c, it takes MUCH longer to start streaming at full speed. If you wait a few minutes after invoking ctrl-c, it starts just fine.  I have had this same behavior with serial connections that are not closed properly, so that is my gut instinct.  I will be looking through the code, but in the mean time I thought I would let you know.

Cheers,
Kel

nicolaslauzier:
I don't know if there is a way to catch ctrl+c inside of a ROS node. If so, it would make sense to close the serial port properly at this moment.

As far as I recall, there is no such mechanism implemented in the node for the FT sensor.

shaun-edwards:
I believe this can be done with a custom SIGINT handler, see [here](http://wiki.ros.org/roscpp/Overview/Initialization%20and%20Shutdown#Custom_SIGINT_Handler)

futureneer:
I mean, couldn't you just put some sort of clean up command after the
while(ros::ok) loop that closes everything down? That seems to be a common
pattern that I have used before.

Is there any command, say in the rq_sensor_com that could be called after the loop?

PR #43, opened 2015-08-11
Changeset by Pierre-Oliver Proulx (robotiq)

lorenz-halt:
Possibly helping with issue #37

### DIFF
--- a/robotiq_force_torque_sensor/src/rq_sensor_state.cpp
+++ b/robotiq_force_torque_sensor/src/rq_sensor_state.cpp
@@ -154,14 +154,27 @@ static void rq_state_start_stream()
  */
 static void rq_state_run()
 {
+	static unsigned char retries = 0;
+
 	rq_com_listen_stream();
 
 	if(rq_com_get_valid_stream() == false)
 	{
+		retries++;
+		printf("invalid stream : retry #%d\n", retries);
+
+		if (retries >= 100)
+		{
 #if defined(_WIN32)||defined(WIN32) //For Windows
-		stop_connection();
+			stop_connection();
 #endif
-		current_state = RQ_STATE_INIT;
+			current_state = RQ_STATE_INIT;
+			retries = 0;
+		}
+	}
+	else
+	{
+		retries = 0;
 	}
 }
 
--- a/robotiq_force_torque_sensor/src/rq_sensor_state.cpp
+++ b/robotiq_force_torque_sensor/src/rq_sensor_state.cpp
@@ -55,7 +55,7 @@ static enum rq_sensor_state_values current_state = RQ_STATE_INIT;
 static INT_8 rq_state_init_com();
 static void rq_state_read_info_high_lvl();
 static void rq_state_start_stream();
-static void rq_state_run();
+static void rq_state_run(unsigned int max_retries);
 
 //////////////////////
 //Function definitions
@@ -65,7 +65,7 @@ static void rq_state_run();
  * \brief Manages the states of the sensor driver
  * \returns -1 if an error occurs, 0 otherwise
  */
-INT_8 rq_sensor_state(void)
+INT_8 rq_sensor_state(unsigned int max_retries)
 {
 	INT_8 ret;
 
@@ -88,7 +88,7 @@ INT_8 rq_sensor_state(void)
 		break;
 
 	case RQ_STATE_RUN:
-		rq_state_run();
+		rq_state_run(max_retries);
 		break;
 
 	default:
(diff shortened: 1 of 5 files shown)


robotiq_force_torque_sensor/src/rq_sensor_state.cpp
```diff
@@ -154,14 +154,27 @@ static void rq_state_start_stream()
  */
 static void rq_state_run()
 {
+	static unsigned char retries = 0;
```
shaun-edwards: If retries always gets set back to zero in the code below, then why declare it as static.  It would be better to declare it as a non-static variable with an initial value of zero.

In addition, I would use a larger size integer.  It would be very easy to specify a number of retries that cannot be represented by the `unsigned char` type.  The compiler may or may not detect such an issue (I'm not sure).

robotiq_force_torque_sensor/src/rq_sensor_state.cpp
```diff
@@ -154,14 +154,27 @@ static void rq_state_start_stream()
...
+
 	rq_com_listen_stream();
 
 	if(rq_com_get_valid_stream() == false)
 	{
+		retries++;
+		printf("invalid stream : retry #%d\n", retries);
+
+		if (retries >= 100)
```
shaun-edwards: The retry limit should be declared as a constant, with a meaningful name either at the top of this function or the top of this source file.

shaun-edwards:
Can you clarify how this addresses #37.  Does simply retrying get the connection "unstuck"?

lorenz-halt:
I don't have any insights into the decision of types by Pierre.
Pierre told me that, if the USB connection breaks down unexpected, the stringstream in corrupted. On my setup, the retries effectively prevent the annoying endless waiting-connection loop, and I can CTRL-C or un- and replug the USB connection.

shaun-edwards:
I still think my comments about the types are valid and easy to address.  

When there is an issue, does the driver print out 100 retry messages before re-initializing?  Does this result in "self-healing" or are the instances where you need to reset the driver to recover?

I'm not excited about it printing out a message for every retry.  Since the message is not a ROS log, it can't be easily silenced.

poproulx:
Declaring the variable inside the function limits its scope to the function only. It gets initialized to 0 only once because it is static.
I agree that the type could be bigger.
I agree that the number of retries should be a constant.
The print instruction is there for debugging purpose. It does not need to be pushed.

When there is a communication issue such as shorting the communication wires, it might take a few data frames before the sensor detects a valid frame. Allowing the state machine to sample more frames add robustness to the communication.

shaun-edwards:
> Declaring the variable inside the function limits its scope to the function only. It gets initialized to 0 only once because it is static.

Declaring the value as static requires resetting it [here](https://github.com/ros-industrial/robotiq/pull/43/files#diff-0afb6d30a3a686d8bbb5fde415959e06R172) and [here](https://github.com/ros-industrial/robotiq/pull/43/files#diff-0afb6d30a3a686d8bbb5fde415959e06R177).  If it were a local variable, initialized to zero, then the two lines above would not be needed.  It will also be clearer code and easier to maintain.  Unless there is a clear reason why the variable value should be maintained between function calls (it keeps it value, it doesn't or it saves memory allocation, not really), declaring it as static is not required.

poproulx:
The value of the variable needs to be kept from call to call. If it was local, it would be initialized every time the function is called and it would always be equal to 1 when testing it. Either it is declared as static global or static local.

shaun-edwards:
OK..now I see it.  My mistake.

shaun-edwards:
Friendly ping....

There was just a few minor modifications required before this can be accepted:
- change retry type to larger type
- number of retries should be a constant and given a better name
- print instruction should be changed to `ROS_DEBUG` call or removed.

ipa-frn:
I'll see what I can do within the next week.

lorenz-halt:
I'll see what I can do within the next week.

shaun-edwards:
@ipa-lth, ping...

robotiq_force_torque_sensor/src/rq_sensor_state.cpp
lorenz-halt: Is there a best practice how to resolve such an issue? This part is somehow unrelated to ros and probably should not include ros.h just to use the logging channel.
I would remove it for now, but the general question stays the same

shaun-edwards: To what issue are you referring?

lorenz-halt: to the print instruction: It should not be a print, but in this file there is no ROS. It's more a library which is then wrapped into a ros node afterwards. So in this file, roslogging is not available, but the output might be worth piping to ros anyway...

shaun-edwards: ROS-I solves this by using logging macros [here](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/simple_message/include/simple_message/log_wrapper.h).  By defining a compiler time macro [here](https://github.com/ros-industrial/industrial_core/blob/indigo-devel/simple_message/CMakeLists.txt#L15), the ROS or non-ROS logger versions are used.

lorenz-halt:
I think that is the best I can do for now, without refactoring the whole thing.
We don't have a working sensor here right now, someone should actually test it before merging!

shaun-edwards:
If anyone has hardware and can test this patch, please respond to this PR.  Otherwise, I will keep this open until someone can test with hardware.

@ipa-lth, thanks for the contribution.

armando-montalvo:
@shaun-edwards I tested it on our sensor and actually found a minor bug. This bug is fixed now.
You may want to merge

lorenz-halt:
@shaun-edwards I tested it on our hardware and it worked fine. I fixed a minor bug in my last commit.

shaun-edwards:
@ipa-lth, thanks for testing. +1 on the merge.

PR #46, opened 2016-01-25
CMakeLists general cleanup for robotiq_force_torque_sensor

progtologist:
Added dependency on message generation for rq_sensor and rq_test_sensor targets, added install targets and made some minor cleanup on CMakeLists.txt

### DIFF
--- a/robotiq_force_torque_sensor/CMakeLists.txt
+++ b/robotiq_force_torque_sensor/CMakeLists.txt
@@ -10,14 +10,18 @@ find_package(catkin REQUIRED COMPONENTS
 
 ## Generate messages in the 'msg' folder
 add_message_files(
-   FILES
-   ft_sensor.msg
+  DIRECTORY
+    msg
+  FILES
+    ft_sensor.msg
 )
 
 ## Generate services in the 'srv' folder
 add_service_files(
+  DIRECTORY
+    srv
   FILES
-  sensor_accessor.srv
+    sensor_accessor.srv
 )
 
 ## Generate added messages and services with any dependencies listed here
@@ -30,7 +34,7 @@ generate_messages(
 ## catkin specific configuration ##
 ###################################
 catkin_package(
-  CATKIN_DEPENDS message_runtime
+  CATKIN_DEPENDS message_runtime std_msgs
 )
 
 ###########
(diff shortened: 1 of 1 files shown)


shaun-edwards:
@gavanderhoorn, can you review these small `CMakeLists` changes.

robotiq_force_torque_sensor/CMakeLists.txt
```diff
@@ -10,12 +10,14 @@ find_package(catkin REQUIRED COMPONENTS
 
 ## Generate messages in the 'msg' folder
 add_message_files(
-   FILES
-   ft_sensor.msg
+  DIRECTORY msg
+  FILES
+  ft_sensor.msg
```
gavanderhoorn: Specifying the directory here should not be necessary (as `msg` is the default), but it doesn't hurt either. Personally I like to be as explicit as possible.

gavanderhoorn: Changing the indent to fit other pkgs would be nice though. Suggestion:

```
add_message_files(
  DIRECTORY
    msg
  FILES
    ft_sensor.msg
)
```

robotiq_force_torque_sensor/CMakeLists.txt
```diff
@@ -10,12 +10,14 @@ find_package(catkin REQUIRED COMPONENTS
...
 )
 
 ## Generate services in the 'srv' folder
 add_service_files(
+  DIRECTORY srv
   FILES
   sensor_accessor.srv
 )
```
gavanderhoorn, line 25: Same comments here.

Could you please update the indentation?

gavanderhoorn:
Only one other comment: @progtologist: can you explain why you want to explicitly list the headers in the `add_executable(..)` statements? Are you using QtCreator?

CMake generated Makefiles should automatically be able to figure out which headers are dependencies of which targets, so that should not be necessary.

gavanderhoorn:
Also: in your commit message you write:

> Added dependency on message generation for [..]

But I cannot find that in 9ad0c256, and the current `CMakeLists.txt` seems to already have that?

progtologist:
@gavanderhoorn I explicitly list the headers for QtCreator, they are not needed in the general case of course, it just makes (my) life easier.
The dependency is added with the "${PROJECT_NAME}_gencpp" in lines 50 and 54.
I can also update the indentation as per your requests. Do you want me to do that and remove the explicit headers inclusion?

gavanderhoorn:
@progtologist wrote:

>  I explicitly list the headers for QtCreator, they are not needed in the general case of course, it just makes (my) life easier.

Yes, I figured that's why you did it.

> The dependency is added with the "${PROJECT_NAME}_gencpp" in lines 50 and 54.

Ah, ok. I was confused because `message_generation` is also a package name.

> I can also update the indentation as per your requests. Do you want me to do that and remove the explicit headers inclusion?

Yes. I don't think we should include editor/tool-specific work-arounds in packages that are intended to be generic. If you could remove it, that would be appreciated (and the `GLOB` as well).

progtologist:
Everything is now ready ;-)

gavanderhoorn:
Looking good.

Could I ask you one more thing? Could you squash your two commits into one? I suggest something like "force_torque_sensor: CMakeLists cleanup." for the commit message (the `robotiq` prefix isn't necessary, as this is a PR against this repository).

That would keep the repository history clean.

progtologist:
Squashed

gavanderhoorn:
@shaun-edwards: I haven't had a chance to test this, but the `CMakeLists.txt` seems ok like this.

shaun-edwards:
+1 @progtologist, thanks for the contribution (and updating things based on @gavanderhoorn's review).

PR #47, opened 2016-02-04
Update rq_sensor_com.cpp

nlauzier:
On some systems, this option was not set properly by default which caused connectivity issues.


gavanderhoorn:
Was this already reviewed (internally fi)?

nicolaslauzier:
Yes, this was reviewed internally at Robotiq and it is a bug that we had on another code base. We are pushing this on the ROS package as it will certainly improve connectivity for all our users. This was a serial port setting that was not defined and it was therefore "floating" (depending on its existing initialization).

PR #51, opened 2016-02-12
Fixed the discrepancy between the visual and collision origins of link_0

kruusamae:
For some reason, the origin tags for link_0 visual and collision were different. This resulted in motion planner thinking there was a collision even when visually there was no collision.

### DIFF
--- a/robotiq_s_model_visualization/cfg/s-model_finger_articulated_macro.xacro
+++ b/robotiq_s_model_visualization/cfg/s-model_finger_articulated_macro.xacro
@@ -20,7 +20,7 @@ finger(i.e. finger_1, etc...).
 				</material>
 			</visual>
 			<collision>	
-				<origin xyz="-0.0455 0.0414 0.036" rpy="0 3.1416 -1.57"/>
+				<origin xyz="0.020 0 0" rpy="0 0 0"/>
 				<geometry>
 					<mesh filename="package://robotiq_s_model_visualization/meshes/s-model_articulated/collision/link_0.STL" />
 				</geometry>


shaun-edwards:
@kruusamae, thanks for the contribution.  Seems as though this would have been caught sooner.  Is it possible for you to show before/after screenshots of the model in rviz (showing the collision).

kruusamae:
Here are the visual and collision models for the robotiq s model before and after the fix. There's no change in visual model, only in the collision model.
robotiq_collision_model_fix

shaun-edwards:
+1...merging.

ibaranov-cp:
Off topic, but @kruusamae how did you get the finger states publishing?

I don't see any method that publishes to /joint_states, or anything similar.

In fact, do you model the object that the fingers would wrap around?

kruusamae:
@ibaranov-cp 
do you mean publishing the actual finger states from robotiq?
as far as i know there isn't anything to do it in the current robotiq repo. in our group, we created a simple publisher that does exactly that and we hope to make a pull request very soon.

ibaranov-cp:
That would be great!

tkkhuu:
Hi I'm using a Robotiq hand and I need to install the driver for indigo following the instruction from the ROS Wiki page: http://wiki.ros.org/robotiq. Then I build using catkin_make and got the errors: 
- robotiq_force_torque_sensor/ft_sensor.h: No such file or directory
- "robotiq_force_torque_sensor/sensor_accessor.h": No such file or directory
- robotiq_c_model_control/CModel_robot_input.h: No such file or directory

Do you have any idea on how to fix these?

PR #54, opened 2016-03-14
Created robotiq_joint_state_publisher package

kruusamae:
Publishes all the Robotiq joint values to joint_states topic. I use it to visualize the actual state of Robotiq gripper in RViz.
Currently supports only Robotiq S-model gripper.

### DIFF
--- a/robotiq_joint_state_publisher/CMakeLists.txt
+++ b/robotiq_joint_state_publisher/CMakeLists.txt
@@ -2,8 +2,6 @@ cmake_minimum_required(VERSION 2.8.3)
 project(robotiq_joint_state_publisher)
 
 ## Find catkin macros and libraries
-## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
-## is used, also find other catkin packages
 find_package(catkin REQUIRED
   roscpp
   roslib
@@ -11,52 +9,16 @@ find_package(catkin REQUIRED
   robotiq_s_model_control
   )
 
-## System dependencies are found with CMake's conventions
-# find_package(Boost REQUIRED COMPONENTS thread)
-
-## Uncomment this if the package has a setup.py. This macro ensures
-## modules and global scripts declared therein get installed
-## See http://ros.org/doc/api/catkin/html/user_guide/setup_dot_py.html
-# catkin_python_setup()
-
-#######################################
-## Declare ROS messages and services ##
-#######################################
-
-## Generate messages in the 'msg' folder
-# add_message_files(
-#   FILES
-#   Message1.msg
-#   Message2.msg
-# )
-
-## Generate services in the 'srv' folder
-# add_service_files(
-#   FILES
-#   Service1.srv
-#   Service2.srv
-# )
-
-## Generate added messages and services with any dependencies listed here
-# generate_messages(
-#   DEPENDENCIES
-#   std_msgs  # Or other packages containing msgs
-# )
-
 ###################################
 ## catkin specific configuration ##
 ###################################
 ## The catkin_package macro generates cmake config files for your package
-## Declare things to be passed to dependent projects
 ## INCLUDE_DIRS: uncomment this if you package contains header files
 ## LIBRARIES: libraries you create in this project that dependent projects also need
 ## CATKIN_DEPENDS: catkin_packages dependent projects also need
 ## DEPENDS: system dependencies of this project that dependent projects also need
 catkin_package(
-#  INCLUDE_DIRS include
-#  LIBRARIES robotiq_joint_state_publisher
-#  CATKIN_DEPENDS other_catkin_pkg
-#  DEPENDS system_lib
+ CATKIN_DEPENDS roscpp roslib sensor_msgs robotiq_s_model_control
 )
 
 ###########
@@ -64,7 +26,6 @@ catkin_package(
 ###########
 
 ## Specify additional locations of header files
-## Your package locations should be listed before other locations
 include_directories(
     ${catkin_INCLUDE_DIRS}
 )
@@ -73,7 +34,6 @@ include_directories(
 add_executable(s_model_joint_states src/s_model_joint_states.cpp)
 
 ## Add cmake target dependencies of the executable/library
-## as an example, message headers may need to be generated before nodes
 add_dependencies(s_model_joint_states robotiq_s_model_control_generate_messages_cpp)
 
 ## Specify libraries to link a library or executable target against
@@ -81,50 +41,3 @@ target_link_libraries(s_model_joint_states
     ${catkin_LIBRARIES}
 )
 
-#############
-## Install ##
-#############
-
-# all install targets should use catkin DESTINATION variables
-# See http://ros.org/doc/api/catkin/html/adv_user_guide/variables.html
-
-## Mark executable scripts (Python etc.) for installation
-## in contrast to setup.py, you can choose the destination
-# install(PROGRAMS
-#   scripts/my_python_script
-#   DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-# )
-
-## Mark executables and/or libraries for installation
-# install(TARGETS motoman_ax motoman_ax_node
-#   ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-#   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
-#   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
-# )
-
-## Mark cpp header files for installation
-# install(DIRECTORY include/${PROJECT_NAME}/
-#   DESTINATION ${CATKIN_PACKAGE_INCLUDE_DESTINATION}
-#   FILES_MATCHING PATTERN "*.h"
-#   PATTERN ".svn" EXCLUDE
-# )
-
-## Mark other files for installation (e.g. launch and bag files, etc.)
-# install(FILES
-#   # myfile1
-#   # myfile2
-#   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
-# )
-
-#############
-## Testing ##
-#############
-
-## Add gtest based cpp test target and link libraries
-# catkin_add_gtest(${PROJECT_NAME}-test test/test_motoman_ax.cpp)
-# if(TARGET ${PROJECT_NAME}-test)
-#   target_link_libraries(${PROJECT_NAME}-test ${PROJECT_NAME})
-# endif()
-
-## Add folders to be run by python nosetests
-# catkin_add_nosetests(test)
--- a/robotiq_joint_state_publisher/CMakeLists.txt
+++ b/robotiq_joint_state_publisher/CMakeLists.txt
@@ -1,0 +1,130 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(robotiq_joint_state_publisher)
+
+## Find catkin macros and libraries
+## if COMPONENTS list like find_package(catkin REQUIRED COMPONENTS xyz)
+## is used, also find other catkin packages
+find_package(catkin REQUIRED
+  roscpp
+  roslib
+  sensor_msgs
+  robotiq_s_model_control
+  )
+
+## System dependencies are found with CMake's conventions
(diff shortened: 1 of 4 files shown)


shaun-edwards:
@kruusamae, thanks for the contribution.  I'll review these changes and get back with you.

robotiq_joint_state_publisher/CMakeLists.txt
```diff
@@ -0,0 +1,130 @@
+cmake_minimum_required(VERSION 2.8.3)
```
shaun-edwards, line 1: Please remove the auto-generated comments from the CMakeLists.txt

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
+//  The MIT License (MIT)
```
shaun-edwards: Would you be willing to re-license this contribution?  The accepted ROS-I licenses are BSD or Apache 2.0 (preferred).

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+/**
+ * Finger class for single Robotiq S-model finger.
+ */
+class Finger {
+ public:
+  Finger() { position = 0; }				///< Default constructor for creating Finger, position is set 0
+  Finger(int pos) { position = pos; }			///< Create Finger, position is @param pos
+  Finger(const Finger &f) { position = f.position; }	///< Create Finger, position is taken form @param Finger
+  inline double joint_1() const;			///< joint_1 value for Finger
```
shaun-edwards: The ROS convention for [methods](http://wiki.ros.org/CppStyleGuide#Function_.2BAC8_Methods) is `camelCase`  These should be named `joint1`, `joint2`, etc..

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  inline double joint_2() const;			///< joint_2 value for Finger
+  inline double joint_3() const;			///< joint_3 value for Finger
+  int position;						///< Position of the Finger
+};
+
```
shaun-edwards, line 61: I'm not familiar with the S-model kinematics/interface.  Can you explain (in a comment) how the joint angles are derived from a single `position`?  I assume these calculation are only correct in a non-interfering state.

There's also a few too many magic numbers for me.  I don't know that making named constants really helps here, since the values seem arbitrary.  Is there a meaning behind the repeated numbers, like `70` and `140`

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  else return -55.0 * DEG_TO_RAD;
+}
+
+/**
+ * Robotiq S-model with three Fingers.
+ */
+class Robotiq3 {
+ public:
+  Robotiq3() { mode = BASIC; scissor = 137; joint_positions.resize(11, 0.0); }		///< Default constructor for creating Robotiq3
+  inline double scissor_joint() const;							///< Joint values for so-called scissor joints
```
shaun-edwards: See method naming comment above.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  if(0 <= position && position <= 140) return (-70.0/140.0 * DEG_TO_RAD) * position;
+  else return -55.0 * DEG_TO_RAD;
+}
+
+/**
+ * Robotiq S-model with three Fingers.
+ */
+class Robotiq3 {
+ public:
+  Robotiq3() { mode = BASIC; scissor = 137; joint_positions.resize(11, 0.0); }		///< Default constructor for creating Robotiq3
```
shaun-edwards: The initialization of the vector can be done via [initializer list](http://en.cppreference.com/w/cpp/language/initializer_list) and the [fill constructor](http://www.cplusplus.com/reference/vector/vector/vector/).  Just an FYI, no need to change.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  int mode;										///< Robotiq grasping mode
+  std::vector<double> joint_positions;							///< All the joint values
+};
+
+/**
+ * Calculate scissor_joint based on the scissor position.
+ */
+inline double Robotiq3::scissor_joint() const {
+  // fully open is +20, fully closed is -10
+  // std::cout << scissor << std::endl;
```
shaun-edwards: Remove commented out code.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+ * Callback function for "SModelRobotInput" topic.
+ * */
+void Robotiq3::callback(const robotiq_s_model_control::SModel_robot_input::ConstPtr &msg) {
+  finger_left = Finger(msg->gPOA);		// set left finger position
+  finger_right = Finger(msg->gPOB);		// set right finger position
+  finger_middle = Finger(msg->gPOC);		// set middle finger position
+  scissor = msg->gPOS;				// set scissor position
+  mode = msg->gMOD;				// set mode
+
+  // if (mode == BASIC)        scissor = 137;
```
shaun-edwards: Should this code be commented out?  If so, are the mode and constant values required?

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+#include <vector>
+#include <string>
+#include <csignal>
+#include <ros/ros.h>
+#include <sensor_msgs/JointState.h>
+#include <robotiq_s_model_control/SModel_robot_input.h>
+#include <robotiq_s_model_control/SModel_robot_output.h>
+
+
+const int BASIC   = 0;
```
shaun-edwards: Should this be an C++ enumeration?  If so, see ROS convention for [enumerations](http://wiki.ros.org/CppStyleGuide#Enumerations)

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+
+  // joint state publisher
+  ros::Publisher joint_pub;
+  joint_pub = nh.advertise<sensor_msgs::JointState>("joint_states", 10);
+
+  // robotiq state message subscriber
+  ros::Subscriber joint_sub;
+  joint_sub = nh.subscribe("SModelRobotInput", 10, &Robotiq3::callback, &robotiq);
+
+  // wait until the URDF has been loaded to avoid publishing joints prematurely
```
shaun-edwards: What happens if you don't wait for the URDF to be loaded?  Under what circumstances does the node get started before the URDF is loaded?

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  while (!ros::isShuttingDown() && !nh.hasParam("/robot_description")) {
+    ROS_ERROR("[s_model_joint_states] Waiting for URDF");
```
shaun-edwards: Should this be a "ROS_WARNING"?

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  }
+
+  // set pseudo-namespace
+  std::string gripper_prefix;
+  if (argc > 1) gripper_prefix = argv[1];
+  else gripper_prefix = "";
+
+  // joint names for sensor_msgs::JointState message
+  // order matters!
+  std::vector<std::string> joint_names(11, "");
```
shaun-edwards: The `joint_names` and `joint_positions` vectors split across the node and the `Robotiq3` class, but they are required to be the same length.  Do you think it would be better to embed the joint names in the `Robotiq3` class?

I'd also recommend using and enumeration for the name/position indices(old school) or perhaps a `std::map`

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  joint_names.at(3).assign(gripper_prefix + "finger_1_joint_3");
+  joint_names.at(4).assign(gripper_prefix + "palm_finger_2_joint");
+  joint_names.at(5).assign(gripper_prefix + "finger_2_joint_1");
+  joint_names.at(6).assign(gripper_prefix + "finger_2_joint_2");
+  joint_names.at(7).assign(gripper_prefix + "finger_2_joint_3");
+  joint_names.at(8).assign(gripper_prefix + "finger_middle_joint_1");
+  joint_names.at(9).assign(gripper_prefix + "finger_middle_joint_2");
+  joint_names.at(10).assign(gripper_prefix + "finger_middle_joint_3");
+
+  // std::cout << "GRIPPER JOINT NAMES: ";
```
shaun-edwards: Remove commented out code.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
```diff
@@ -0,0 +1,208 @@
...
+  // std::copy(joint_names.begin(), joint_names.end(), std::ostream_iterator<std::string>(std::cout, "\n"));
+
+  // Output JointState message
+  sensor_msgs::JointState joint_msg;
+  // Joint names to JointState message
+  joint_msg.name = joint_names;
+
+  while (!ros::isShuttingDown()) {
```
shaun-edwards: I normally use `ros::ok()`, reads better.  Is there a reason to use `isShuttingDown` (just curious, don't have to fix).

shaun-edwards:
@kruusamae, thanks for the contribution.  Looks pretty good overall.  A few issues, but if you can address them, I'd be happy to accept the PR.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
kruusamae: You're right! Mode and constants are not required.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
kruusamae: As far as I can tell, nothing happens. This wait was useful for debugging. I removed it now.

robotiq_joint_state_publisher/src/s_model_joint_states.cpp
kruusamae: I guess it makes sense  to have `joint_names` in `Robotiq3` class, so I did that. But since `sensor_msgs::JointState` also has separate vectors for names and positions, I can't see any benefits from using a `std::map`.

kruusamae:
@shaun-edwards 
Thank you for your feedback!
If you have any other comments, just let me know.

shaun-edwards:
@kruusamae, thanks for the updates.  I believe the MIT license is still in `s_model_joint_states.cpp`.  Did you mean to leave it in?

kruusamae:
@shaun-edwards 
Yeah, that was intentional. I am no expert on software licensing but since the underlying work was licensed under MIT, it seemed like I am required to preserve the original notice. I don't know, maybe I'm too cautious or just wrong, but the way I see it, is that I can't simply overwrite the original license. All the changes that I've made are under BSD. If you have better knowledge on the topic and/or some other suggestions, please let me know. Because as I said, I am no expert.

shaun-edwards:
@kruusamae, I'm not a lawyer...but.....If all of the code comes from The University of Texas at Austin students and employees, then The University of Texas at Austin is free to change the terms of the license, as they see fit.  If the code was contributed by someone else, then you would have to get their permission to change the license.  If you feel you can change the license, then replace the MIT license with the BSD version.  If you can't change the license, then remove the BSD terms.  It would be confusing to have two license for the same code practically speaking, legally I'm not even sure what that means.

kruusamae:
@shaun-edwards, OK, I talked it through and removed the MIT license. Sorry about the confusion before.

shaun-edwards:
+1...merging.  Thanks for the contribution @kruusamae.

PR #59, opened 2016-04-15
change prefix to optional rosparameter

Jntzko:
Starting the node via a launch file produced an error:
The joint names were changed to e.g. "__name:=nodenamepalm_finger_joint_1".
To fix this, the commandline argument was changed into an rosparameter.
If you started the node via rosrun:
rosrun robotiq_joint_state_publisher s_model_joint_states "gripper_prefix"
Now you start it with:
rosrun robotiq_joint_state_publisher s_model_joint_states _prefix:="gripper_prefix"

### DIFF
--- a/robotiq_joint_state_publisher/README.md
+++ b/robotiq_joint_state_publisher/README.md
@@ -11,5 +11,5 @@ Run:
 
 or
 
-`rosrun robotiq_joint_state_publisher s_model_joint_states <gripper_prefix>`
+`rosrun robotiq_joint_state_publisher s_model_joint_states _prefix:=<gripper_prefix>`
 
--- a/robotiq_joint_state_publisher/src/s_model_joint_states.cpp
+++ b/robotiq_joint_state_publisher/src/s_model_joint_states.cpp
@@ -177,19 +177,19 @@ inline std::vector<std::string>  Robotiq3::jointNames() {
  * Main method.
  */
 int main(int argc, char *argv[]) {
-
-  // set user-specified prefix
-  std::string gripper_prefix;
-  if (argc > 1) gripper_prefix = argv[1];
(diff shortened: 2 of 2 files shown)


shaun-edwards:
Do you know why the command line argument approach doesn't work?

Jntzko:
The command line argument does work, the problem is when you start the node within a launch file. Roslaunch set the "__name" parameter behind the arguments, so when no other arguments were specified, the name parameter is the first argument. That means all the joint names look like "nodenamelink1_link2".

v4hn:
:+1: Please merge.
An alternative bugfix would only move the `ros::init` call above `if(argc > 1) ...`.
However this would leave us with the following roslaunch syntax to invoke the node

```
<node name="passive_joint_publisher" pkg="robotiq_joint_state_publisher" type="s_model_joint_states" args="s_model_" />
```

compared to the (imho) more readable

```
<node name="passive_joint_publisher" pkg="robotiq_joint_state_publisher" type="s_model_joint_states">
   <param name="prefix" value="s_model_" />
</node>
```

as the node is new in this repository, it should not (yet) be an issue to change the API.

shaun-edwards:
@kruusamae, do you agree with @v4hn?  I'm leaning in @v4hn's direction, so unless you object, I plan to merge.

kruusamae:
@shaun-edwards, at this moment I don't have a strong preference one way or the other. I guess using rosparam makes sense. Overall, I think this is one of those things that should be consistent throughout the whole robotiq package - maybe some guidelines telling developers how and when to use rosparam and argv?

shaun-edwards:
+1 for this change.  Sorry for the delay.  Thanks @Jntzko for the contribution and thanks @v4hn and @kruusamae for the comments.

PR #60, opened 2016-04-21
High resolution model ft-sensor

armando-montalvo:
*Adding high resolution model of the gripper.
*Adding .stl model from sensor (taken from CAD data).


v4hn:
+1 please merge
@armando-montalvo: Having a high-resolution mesh is great. I'd prefer to also have a low-resolution _mesh_ for collision checking. The current cylinder ignores the expansion+cable at one end, so we had to increase the radius quite a bit to get no collision in practice (8.5cm). However, this restricts the overall setup quite a bit. Could you also add a low-resolution mesh of the sensor and use it for the collision geometry?

shaun-edwards:
+1 to low resolution collision model.  In the past we have used convex hulls to be safe.  Simply reducing the resolution could result in collisions depending on the simplification algorithm.

Where did you get the CAD models?  Are they publicly available?

Is there a reason not to replace the existing URDF with this one?  Is it concern for backwards compatibility?

gavanderhoorn:
@shaun-edwards wrote:

> Is there a reason not to replace the existing URDF with this one?

It might be nice to follow the ros-i convention of having two sub dirs of `meshes`, for storing detailed meshes (`visual`) and lower quality meshes (`collision`). The xacro macro could then refer to these in the related `link` sub-elements. No need for separate urdfs then.

armando-montalvo:
@v4hn: you mean to make a model with primitives that somehow resembles better the form of the sensor? Another idea could be to reduce a lot the quality of the stl file.
@shaun-edwards : yes, the CAD model is in the webpage of the sensor: http://robotiq.com/products/robotics-force-torque-sensor/ (You can see it under Download Links). I just converted it to .stl with Blender and reduced its quality a bit.

v4hn:
On Tue, Apr 26, 2016 at 12:37:30AM -0700, Armando Montalvo wrote:

> @v4hn: you mean to make a model with primitives that somehow resembles better the form of the sensor?

No. As @shaun-edwards described a mesh of the convex hull would be nice.
Primitives only get you this far.

shaun-edwards:
@armando-montalvo could you make the following changes?
- Use convex hulls for collision models as @v4hn suggests
- follow @gavanderhoorn suggestion to follow the ROS-I convention and not separate the URDFs?

shaun-edwards:
Included in #64 (merged)

PR #61, opened 2016-04-28
Correct joint limits for s Model

Jntzko:
The upper and lower limits for the joints were identified with the
robotiq_joint_state_publisher for the s model by moving to different poses via the SModelSimpleController.py.
This was necessary in order to use moveit with this urdf out of the box.

### DIFF
--- a/robotiq_s_model_visualization/cfg/s-model_articulated_macro.xacro
+++ b/robotiq_s_model_visualization/cfg/s-model_articulated_macro.xacro
@@ -40,21 +40,19 @@ there are multiple hands then a prefix followed by an "_" is needed.
 		<!-- end of link list -->
 		<!-- joint list -->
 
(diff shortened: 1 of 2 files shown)


shaun-edwards:
Can you clarify what you mean by "using" MoveIt?  Are you planning for the articulated fingers?

Jntzko:
I've written an "GripperCommand" action server for the gripper that supports specific open and close positions (https://github.com/TAMS-Group/robotiq_s_model_action_server) and added it to moveit's list of controllers. When using moveit's planning capability to go to these states I get errors from moveit that some joints are out of their joint limits, so I adjusted all joint limits to their respective maximum/minimum. I got the values by moving the hand around with the SModelSimpleController.py.

shaun-edwards:
@Jntzko, thanks for the contribution.  +1...merging.

PR #62, opened 2016-05-02
Toolframe for robotiq's s model gripper

Jntzko:
New link named toolframe.
The toolframe has its origin between the three fingers.

The toolframe link is handy to use as an end effector to plan for an arm
with the gripper attached.

### DIFF
--- a/robotiq_s_model_visualization/cfg/s-model_articulated_macro.xacro
+++ b/robotiq_s_model_visualization/cfg/s-model_articulated_macro.xacro
@@ -31,6 +31,7 @@ there are multiple hands then a prefix followed by an "_" is needed.
 				</material>
 			</collision>
 		</link>
+		<link name="${prefix}tool0"/>
         
         <inertial>
             <origin xyz="0 0 0" rpy="0 0 0"/>
@@ -59,6 +60,11 @@ there are multiple hands then a prefix followed by an "_" is needed.
 			<child link="${prefix}finger_middle_link_0"/>
 			<axis xyz="0 0 1"/>
 			<origin xyz="0.0455 0.0214 0" rpy="0 0 1.57"/>
+		</joint>
+		<joint name="${prefix}palm_tool0" type="fixed">
+			<parent link="${prefix}palm"/>
+			<child link="${prefix}tool0"/>
+			<origin xyz="0.0 0.0875 0.0" rpy="0 0 1.5707"/>
 		</joint>
 		<!-- end of joint list -->
 	</xacro:macro>


shaun-edwards:
How did you select the location of the `toolframe`?  Does it make sense for everyone to use the same frame or should the tool frame be specified for each application (via combination by xacro)?

Jntzko:
The position of the toolframe is a compromise for the different modes. For the pinch mode the toolframe should obvious be in the centre between the fingertips, but it should also work with this position. For the basic and wide mode it is not really clear where the toolframe should be, but this solution works pretty well for these two modes. To use the scissor mode with this toolframe you have to apply an additional transformation. 
I think it is not a good idea to specify it in the xacro, because you can change the mode at runtime, but you can not change the toolframe transformation. A more appropriate solution would set the toolframe at runtime depending on the current mode but this probably does not work well with move it... so this is a simple compromise.

shaun-edwards:
> I think it is not a good idea to specify it in the xacro, because you can change the mode at runtime, but you can not change the toolframe transformation. 

To clarify, I meant that we should allow users to specify tool frames because there isn't an obvious frame.  It's possible for users to do this by using a xacro that includes a robotiq and a tool frame transform separately.

> A more appropriate solution would set the toolframe at runtime depending on the current mode but this probably does not work well with move it... so this is a simple compromise.

This bugs me about MoveIt to no end.  We've made several work-arounds that utilize TF to calculate poses from "variable" tool frames to one MoveIt understands.  However, we've never figured out how to fix this in MoveIt itself.

v4hn:
@shaun-edwards Could you clarify your idea of toolframe-support for this gripper?
Don't specify a toolframe and leave it to the user to specify one in the xacro that includes your gripper definition?
This is _always_ possible, independent of this request.
This request aims to have a "simple" toolframe around that works for many scenarios out of the box,
reducing the entry barrier.
Is this enough motivation to get this change upstream?

> This bugs me about MoveIt to no end. We've made several work-arounds that utilize TF to calculate poses from "variable" tool frames to one MoveIt understands. However, we've never figured out how to fix this in MoveIt itself.

I might have to look into this code anyway over the next months.
Could you specify a bit more in detail, what _exactly_ you would like to see supported by moveit?

shaun-edwards:
@Jntzko and @v4hn, I feel you've out-voted me on this, plus I buy the usability argument.  My one request is to change the frame name to the [ROS-I standard](http://wiki.ros.org/Industrial/Tutorials/Create%20a%20URDF%20for%20an%20Industrial%20Robot#Additional.2BAC8-Standard_Frames) `${prefix}tool0`

shaun-edwards:
@v4hn 

> Could you specify a bit more in detail, what exactly you would like to see supported by moveit?

We tried to implement a MoveIt adapter for our Descartes project that allowed the user to specify a toolpoint pose in an [arbitrary frame](https://github.com/ros-industrial-consortium/descartes/blob/indigo-devel/descartes_moveit/src/moveit_state_adapter.cpp#L120-L128) (relative to the robot base frame) for an [arbitrary toolpoint](https://github.com/ros-industrial-consortium/descartes/blob/indigo-devel/descartes_moveit/src/moveit_state_adapter.cpp#L99-L106)(relative to the robot tool).  This capability is built into every robot controller, but missing from MoveIt as I understand it.  It seems MoveIt does not have access to the full TF, rather it only knows what is in the URDF.  We achieved partial functionality by utilizing TF to get back to a URDF frame.  I think this is a valid approach and would enable things like moving robot based and tool changers.

Jntzko:
I changed the tooframe name as requested to `${prefix}tool0`.

shaun-edwards:
+1..Awesome!...Thanks for the contribution.  Please keep 'em coming.

PR #64, opened 2016-05-04
proper meshes for ft-sensor (based on #60)

v4hn:
This implements the improvements discussed in #60
and also cleans up some stuff around it.


shaun-edwards:
+1.  Thanks for the contribution @v4hn and @armando-montalvo.  Sorry for taking so long to merge.  I am trying to move faster on these things.

PR #65, opened 2016-05-16
Issue prevents RVIZ from loading current STL files

cpaxton:


gavanderhoorn:
Is this to fix a "ascii vs binary stl and `solid` in the header" problem?

cpaxton:
That was the goal, yeah.

gavanderhoorn:
So why close it? :)

cpaxton:
I'm not sure why the integration is failing or the meshes won't show here on github. It was a pretty minor change: I just got rid of the "solid" and replaced with a "robot" tag.

gavanderhoorn:
Travis failing is a configuration issue with the repository, not your fault.

> [..] or the meshes won't show here on github.

What do you mean?

shaun-edwards:
@gavanderhoorn, the file's changed tab doesn't show the model diff.  The diff is likely empty since there isn't a geometry difference.

@cpaxton, I think this is worth merging if the models don't load in Rviz (that's the whole point of the models).

v4hn:
We had no problem with these meshes for the last year, i.e. they show(and showed) up in RViz.
Hence, I'm not sure what this request is supposed to fix.

@gavanderhoorn could you describe and provide a picture of the problem?

shaun-edwards:
@cpaxton, is this still an issue or can this PR be closed.  @v4hn's comment seems to indicate that it isn't an issue.

cpaxton:
Closed the issue. It may be a quirk of one of our platforms. We use a lot of custom code, including custom RVIZ forks and ros-hydro. If I notice it in vanilla rviz in ros-indigo I will provide more documentation.

PR #66, opened 2016-05-27
Changed gripper color to gray

Jntzko:
The color is changed to gray, because the interactive marker for the goal state in rviz is already green and this results to misunderstandings.
I didn't use black because the contours of the gripper are not visible in black.


shaun-edwards:
Can you embed before and after screenshots from rviz into this PR?

Jntzko:
This is the gripper in green:
green
This, as I wanted in gray:
gray
And this in black:
black

shaun-edwards:
@Jntzko thanks for the contribution.  Sorry it took so long to merge (I thought I merged it a while back).  +1 to this...merging...

Jntzko:
This is still missing in jade devel. Would you mind cherry picking it there?

shaun-edwards:
Thanks for the heads up. I plan to sync the two versions in the next couple
days.

Issue #70, opened 2016-06-03
robotiq_s_model_articulated_gazebo_plugins and robotiq_s_model_articulated_msgs haven't been ported forward from groovy-devel


UltronDestroyer:
Do you mind if I make a PR for this? Any reason for holding it back from Hydro, Indigo, Jade?

PR #71, opened 2016-06-03
Bringing forward old changes, an urdf bug fix.

UltronDestroyer:
-Bringing forward Gazebo plugin work, an articulated, underactuated gripper from Groovy devel
- fixing the prefix eval in force torque sensors

Please see the code notes please. Also, referencing #70

### DIFF
--- a/robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
+++ b/robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
@@ -1,0 +1,888 @@
+/*
+ * Copyright 2014 Open Source Robotics Foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+*/
+/*
+    This file has been modified from the original, by Devon Ash
+*/ 
+
+#include <robotiq_s_model_articulated_msgs/SModelRobotInput.h>
+#include <robotiq_s_model_articulated_msgs/SModelRobotOutput.h>
+#include <ros/ros.h>
+#include <string>
+#include <vector>
+
+/*
+
+Due to necessity, I had to change the PID.hh file's definition from private members to public to allow public access of its members. They're private in 1.9 and getter functions aren't implemented until gazebo 3.0. I thought that was silly, and so I hacked around it. The functions directly access the private members by necessity. It can only change if Gazebo patches 1.9 and 2.2 to include getters for it.
+
+I'm not sure exactly where the dependency chain includes PID.hh for the first time, so I've encapsulated all of the gazebo includes. Not pretty, but it works. If you're reading this and know of a better soln', feel free to change it.
+
+*/
+#define private public 
+#include <gazebo/common/Plugin.hh>
+#include <gazebo/common/Time.hh>
+#include <gazebo/math/Angle.hh>
+#include <gazebo/physics/physics.hh>
+#include <robotiq_s_model_articulated_gazebo_plugins/RobotiqHandPlugin.h>
+#undef private
+
+// Default topic names initialization.
+const std::string RobotiqHandPlugin::DefaultLeftTopicCommand  =
+  "/left_hand/command";
+const std::string RobotiqHandPlugin::DefaultLeftTopicState    =
+  "/left_hand/state";
+const std::string RobotiqHandPlugin::DefaultRightTopicCommand =
+  "/right_hand/command";
+const std::string RobotiqHandPlugin::DefaultRightTopicState   =
+  "/right_hand/state";
+
+////////////////////////////////////////////////////////////////////////////////
+RobotiqHandPlugin::RobotiqHandPlugin()
+{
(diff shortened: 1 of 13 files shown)


shaun-edwards:
@TheDash, thanks for this.  I'm in favor of merging this, but I'd like to figure out why this was missed (most likely my fault).  These same set of changes will likely need to be merged into `indigo-devel`.

UltronDestroyer:
Maybe it was my fault for pushing to Groovy. Not sure. Anyways, we can merge it into Jade. I'm about to make a PR soon adding the FT300 model to the ft sensor description.

athackst:
@TheDash would it be possible to remove the whole "side" notion from the gazebo plugin?  It looks like it is enforcing some assumptions on the prefix/naming etc that is not necessarily true.

Also, it looks like the input/output messages have been copied and renamed to their own package so now we will have 2 of essentially the same thing.  I recommend picking one.

UltronDestroyer:
@athackst Can you comment on the locations in the code you're talking about? I'm not too familiar with this code base as it was just an integration project. 

We can definitely use existing messages barring they're internally the same and maintains similar functionality, I'm impartial.

robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
```diff
@@ -0,0 +1,888 @@
...
+
+////////////////////////////////////////////////////////////////////////////////
+void RobotiqHandPlugin::Load(gazebo::physics::ModelPtr _parent,
+                             sdf::ElementPtr _sdf)
+{
+  this->model = _parent;
+  this->world = this->model->GetWorld();
+  this->sdf = _sdf;
+
+  if (!this->sdf->HasElement("side") ||
```
athackst, line 87: @TheDash This is where the implementation of "side" and assignment of static prefixes begins.  I think just requiring "prefix" would be better since that is how it is defined in the xacro

robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
```diff
@@ -0,0 +1,888 @@
...
+    this->jointStates.position[i] = 0;
+    this->jointStates.velocity[i] = 0;
+    this->jointStates.effort[i] = 0;
+  }
+  gzlog << "Initialized the joint state vector" << std::endl;
+
+  // Default ROS topic names.
+  std::string controlTopicName = this->DefaultLeftTopicCommand;
+  std::string stateTopicName   = this->DefaultLeftTopicState;
+  if (this->side == "right")
```
athackst, line 127: @TheDash There is also this notion of setting topic names based on "side".   Control topic name and state topic names should either be specified or match the default from running the controller with a single gripper if you really want a default.

robotiq_s_model_articulated_msgs/msg/SModelRobotInput.msg
```diff
@@ -0,0 +1,154 @@
+# gACT : Initialization status, echo of the rACT bit (activation bit).
```
athackst, line 1: This is the input message that matches the one found in robotiq_s_model_control/msg.  It functionally should work the same.  I actually think this version is better and it might be better to be in it's own package to make the dependency tree smaller if you only care about gazebo and don't want to install the control package.  Assuming of course that this ever gets released as a debian.  However keeping this version and removing the one in robotiq_s_model_control will affect that package and anyone else using it, which would be disruptive.

We can let the maintainers decide how to handle it.

robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
UltronDestroyer, line 87: Agreed. Looks to be a namechange

robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
UltronDestroyer, line 127: It should just append the prefix.

robotiq_s_model_articulated_msgs/msg/SModelRobotInput.msg
UltronDestroyer, line 1: > We can let the maintainers decide how to handle it.

Ok.

robotiq_s_model_articulated_gazebo_plugins/src/RobotiqHandPlugin.cpp
athackst, line 127: Sure.  The default topic names if you launched the hardware controller are /gripper/output = "command" and /gripper/input = "state".  You could append prefix so that it would become "/right_gripper/output" if "right_" was the prefix for the default topic name.

robotiq_s_model_articulated_msgs/msg/SModelRobotInput.msg
shaun-edwards, line 1: It's best practice for messages to be in their own package in order to prevent circular dependencies.  Deleting the `robotiq_s_moedel_control` message will break things.  We generally like to deprecate in one version and then remove in the next.

shaun-edwards:
@TheDash, @athackst, sorry for the delay on my end.  I'm going to merge this change into `indigo-devel` (if needed) as is, since it was present in groovy.

The items that @athackst pointed out above can be fixed in `jade-devel` if we think they are needed.  We don't use gazebo so I'd rather rely on your guys judgement.

wxmerkt:
Hey, what's the status of this PR and can we help/assist in getting it merged?
Thanks,
Wolfgang

shaun-edwards:
@wxmerkt, Thanks for the ping.  I merged #71 and will sync the jade branch soon.

shaun-edwards:
Merged in #71, closing.

wxmerkt:
Awesome thanks @shaun-edwards :)

PR #73, opened 2016-06-14
Add queue_size to SModelRobotInput publisher

Jntzko:
The publisher should be created with an explicit keyword argument
'queue_size'. The queue_size is set to 1, because every message contains
all needed informations.

### DIFF
--- a/robotiq_s_model_control/nodes/SModelTcpNode.py
+++ b/robotiq_s_model_control/nodes/SModelTcpNode.py
@@ -62,7 +62,7 @@ def mainLoop(address):
     rospy.init_node('robotiqSModel')
 
     #The Gripper status is published on the topic named 'SModelRobotInput'
-    pub = rospy.Publisher('SModelRobotInput', inputMsg.SModel_robot_input)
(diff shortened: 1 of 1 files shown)


shaun-edwards:
@Jntzko, thanks for the contribution.  Is there a reason a publisher needs to have the queue size set?  Do you know what the default behavior is if it's not set?

gavanderhoorn:
@shaun-edwards: the requirement (or at least: the display of the warning) for `queue_size` to be set explicitly was introduced with Indigo. See [rospy/Overview/Publishers and Subscribers - Publishing to a topic - queue_size: publish() behavior and queuing](http://wiki.ros.org/rospy/Overview/Publishers%20and%20Subscribers#queue_size:_publish.28.29_behavior_and_queuing) for a good rationale.

+1 to the fact that this is needed. The size of the queue might need some tweaking, but see the wiki link I posted.

shaun-edwards:
@Jntzko, thanks for the contribution.  @gavanderhoorn, thanks for the link/info.  I learned something new.  Merging...

PR #75, opened 2016-06-25
Added travis config file for continuous integration testing

shaun-edwards:
Creating this PR for `indigo-devel` right now.  Will sync up `jade-devel` branch after this is accepted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+sudo: required 
+dist: trusty 
+language: generic 
+compiler:
+  - gcc
+notifications:
+  email:
+    on_success: always
+    on_failure: always
+#    recipients:
+#      - jane@doe
+env:
+  matrix:
+    - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - ROS_DISTRO=kinetic
+matrix:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+    on_failure: always
+#    recipients:
+#      - jane@doe
+env:
+  matrix:
+    - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - USE_DEB=true  ROS_DISTRO="indigo" ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
+    - ROS_DISTRO=kinetic
```
gavanderhoorn, line 18: this will run Kinetic builds in addition to Indigo and Jade?

shaun-edwards, line 18: I'm not sure what the convention is, but with `robotiq`, I'd like to run _all_ the later ros versions that are available, but only require the current version to pass.  This should give us a heads up when future versions are going to fail.

.travis.yml
```diff
@@ -0,0 +1,28 @@
...
+matrix:
+  allow_failures:
+    - env: USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros/ubuntu
+    - env: USE_DEB=true  ROS_DISTRO="jade"   ROS_REPOSITORY_PATH=http://packages.ros.org/ros-shadow-fixed/ubuntu
```
gavanderhoorn, line 22: Any specific reason to allow failures for Jade?

Seeing as you have a `jade-devel` branch now?

gavanderhoorn:
Besides the allowed failures for Jade (which I think are not necessary anymore?), +1.

.travis.yml
shaun-edwards, line 22: I'd like to have a LTS/Indigo version that can build on multiple ROS versions.  That way Jade/Kinetic could be more experimental.  Might be a pipe dream since, robotiq wont' build on anything outside of indigo currently.

PR #76, opened 2016-06-26
Merging PR #71 into indigo

shaun-edwards:


shaun-edwards:
@TheDash, same as #71, but for `indigo-devel`.  Can you review.  I had to add a dependency to `robotiq_s_model_articulated`.  Isn't TravisCI awesome?  

#75 must be merged first (travis CI)

After this is accepted, I will sync `jade-devel`

shaun-edwards:
@TheDash, friendly ping.  Will you be able to review this change?

UltronDestroyer:
@shaun-edwards  My bad. I reviewed it but didn't say anything.

So now that these are in the proper branch. We have another issue that is opened on another ticket: installing this stack via source by default you have to also install gazebo debs or source which is massive (unless you move the gazebo packages out of the ws). My suggestion was to change it to logical dependency separation.

PR #85, opened 2016-08-22
Fixed STL solid/binary warnings on load

### LINKED ISSUE #84: Malformed STL Warnings on S-Model gripper RViz Load
I receive the following warnings when loading the STL files for the S-model gripper:

```
[ WARN] The STL file 'package://robotiq_s_model_visualization/meshes/s-model_articulated/visual/link_0.STL' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
[ WARN] The STL file 'package://robotiq_s_model_visualization/meshes/s-model_articulated/visual/link_1.STL' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
[ WARN] The STL file 'package://robotiq_s_model_visualization/meshes/s-model_articulated/visual/link_2.STL' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
[ WARN] The STL file 'package://robotiq_s_model_visualization/meshes/s-model_articulated/visual/link_3.STL' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
[ WARN] The STL file 'package://robotiq_s_model_visualization/meshes/s-model_articulated/visual/palm.STL' is malformed. It starts with the word 'solid', indicating that it's an ASCII STL file, but it does not contain the word 'endsolid' so it is either a malformed ASCII STL file or it is actually a binary STL file. Trying to interpret it as a binary STL file instead.
```

I believe this is an easy fix, just open the files in MeshLab and re-save them.

Kukanani:
I used MeshLab to re-open and save all of the visual files for the S-Model gripper. Some of the meshes changed as seen below, however, this was only because I also used MeshLab to remove duplicate vertices. The overall volume of the visual models is the same, as can be seen when using the Revision Slider.

Fixes #84. This is the problem addressed by #65, and I'm not sure why that PR was ever accepted. I've always been taught to treat warnings as errors.


gavanderhoorn:
> Fixes #84. This is the problem addressed by #65, and I'm not sure why that PR was ever accepted. 

Afaict, #65 was never merged?

Kukanani:
Looks like #65 prevented loading for a user with a special case. This issue does not prevent loading in the general case, but causes lots of warnings to be printed to console. Can this be merged?

shaun-edwards:

PR #88, opened 2016-12-22
Updated travis for jade builds.  Added rosinstall for soem source dep…

shaun-edwards:
…endency until soem binary is released


jproberge (changes requested):
I know this is an old PR, but the Travis files update was relevant. However, this PR does not pass the Travis checks because of the modifications done to the robotiq_ethercat/CMakeLists.txt and robotiq_ethercat/src/ethercat_manager.cpp

On the other hand, the modifications to the travis.yml and the addition of the rosinstall file are valid. One way to pass the Travis checks, would simply be to dismiss the modifications done to CMakeLists.txt and ethercat_manager.cpp .

PR #91, opened 2017-01-04
Moved publishers and service for force_torque_sensor to private namespace

NikolasE:
So far, the publishers and the service have been in the global namespace. This does not work for multiple sensors and also make configuring interfaces harder as you have to pass the name of the publishers and the service if the node should be able to use both. With the current version, an interface node only gets the node name and can either add the names or use topictools to find the correct topic names.

### DIFF
--- a/robotiq_force_torque_sensor/nodes/rq_sensor.cpp
+++ b/robotiq_force_torque_sensor/nodes/rq_sensor.cpp
@@ -173,9 +173,10 @@ int main(int argc, char **argv)
 		wait_for_other_connection();
 	}
 	
-	ros::Publisher sensor_pub = n.advertise<robotiq_force_torque_sensor::ft_sensor>("robotiq_force_torque_sensor", 512);
-	ros::Publisher wrench_pub = n.advertise<geometry_msgs::WrenchStamped>("robotiq_force_torque_wrench", 512);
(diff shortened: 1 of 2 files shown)


shaun-edwards:
@NikolasE, thanks for the contribution.  Is there a reason that node namespacing with the `ns` argument in a launch file cannot be used to enable multiple sensors?  See [here](http://wiki.ros.org/roslaunch/XML/node#Attributes).

NikolasE:
This would be a workaround and would lead to long names. The names "robotiq_force_torque_sensor" and "robotiq_force_torque_wrench" look like the author already had some kind of namespace in mind. 
I think it's better if the default behaviour would already provide the namespace instead of having to add a parameter. 

Using multiple sensors is a different issue, currently, the driver just tries all /dev/* so that there is no way to select a specific sensor.

shaun-edwards:
The default behavior in ROS is to list services and topics in the node's namespace.  If the number of topics/services is large, then I have seen some namespace hierarchy utilized.  I've also seen "utility" topics and services put in the private node namespace.  I'm not sure the robotiq package falls into either of these categories, so we shouldn't divert from the default behavior (IMHO).

My other hesitation would be backwards compatibility.  Changing topic/service names can cause pain for others.

Issue #92, opened 2017-01-04
Strings as commands for FT300


NikolasE:
The current way to send commands to the FT300 is via strings in a service. To zero the sensor, one has to send a "SET ZRO" (Not to be confused with "SET ZERO"). This leads also to parsing code like 

`	strncpy(get_or_set, &buff[0], 3);
	strncpy(nom_var, &buff[4], strlen(buff) -3);`

What do you think about a srv format like this:

```
uint8 COMMAND_GET_SERIAL_NUMBER=1
uint8 COMMAND_GET_FIRMWARE_VERSION=2
uint8 COMMAND_GET_PRODUCTION_YEAR=4
uint8 COMMAND_SET_ZERO=8
uint8 command
---
bool success
string res
```

This leads to much cleaner code on both sides and the client does not have to remember string-commands. 

(I volunteer to implement the changes)

shaun-edwards:
@NikolasE, thanks for the suggestion.  I think this would indeed be an improvement.  With this kind of change, we have to be careful to think about backward compatibility.  We wouldn't want to break any code that is currently working with the string based service.  I believe there are few options:

- Support both the new and old services, simultaneously (2 service servers with different names).  Add a deprecated warning to the old string based service
- Support either string based arguments or integer commands in a single service.  Add a deprecated warning whenever the string argument isn't empty.
- Add a deprecated warning to jade and then replace the service in kinetic.

What do you think?  I think the second option may be the most user friendly and maintainable.

NikolasE:
The second option looks ok. What do you think about a ROS_WARN if the command-string is not empty? 
```

uint8 COMMAND_GET_SERIAL_NUMBER=1
uint8 COMMAND_GET_FIRMWARE_VERSION=2
uint8 COMMAND_GET_PRODUCTION_YEAR=4
uint8 COMMAND_SET_ZERO=8
uint8 command_id
string command  # deprecated, please use command_id with a value of COMMAND_*
---
bool success
string res
```

shaun-edwards:
Addressed in #93

PR #93, opened 2017-01-05
Replaced string-based interface with numerical constants

NikolasE:
…so that a user does not have to remember commands like 'GET PYE'. String-commands are still supported but generate a ROS_WARN

https://github.com/ros-industrial/robotiq/issues/92

### DIFF
--- a/robotiq_force_torque_sensor/nodes/rq_sensor.cpp
+++ b/robotiq_force_torque_sensor/nodes/rq_sensor.cpp
@@ -50,11 +50,53 @@
 #include "robotiq_force_torque_sensor/ft_sensor.h"
 #include "robotiq_force_torque_sensor/sensor_accessor.h"
 
+typedef robotiq_force_torque_sensor::sensor_accessor::Request Request;
+
 static void decode_message_and_do(INT_8 const  * const buff, INT_8 * const ret);
+static void decode_message_and_do(robotiq_force_torque_sensor::sensor_accessor::Request& req,
+								  robotiq_force_torque_sensor::sensor_accessor::Response& res);
+
 static void wait_for_other_connection(void);
 static int max_retries_(100);
 
 ros::Publisher sensor_pub_acc;
+
+/**
+ * @brief decode_message_and_do Decode the message received and do the associated action
+ * @param req
+ * @param res
+ */
+static void decode_message_and_do(robotiq_force_torque_sensor::sensor_accessor::Request& req,
+								  robotiq_force_torque_sensor::sensor_accessor::Response& res)
+{
+	if (req.command_id == Request::COMMAND_SET_ZERO)
+	{
+		ROS_INFO("Zeroing Sensor");
+		rq_state_do_zero_force_flag();
+		res.success = true;
+		res.res = "Done";
+		return;
+	}
+
+	if (req.command_id != Request::COMMAND_GET_FIRMWARE_VERSION &&
+		req.command_id != Request::COMMAND_GET_PRODUCTION_YEAR &&
+		req.command_id != Request::COMMAND_GET_SERIAL_NUMBER)
+	{
+		ROS_WARN("Unsupported command_id '%i', should be in [%i, %i, %i]",
+			 req.command_id,
+			 Request::COMMAND_GET_SERIAL_NUMBER,
+			 Request::COMMAND_GET_FIRMWARE_VERSION,
+			 Request::COMMAND_GET_PRODUCTION_YEAR);
+		res.success = false;
+		res.res = "Unsupported command_id";
+		return;
+	}
+
+	INT_8 buffer[512];
+	res.success = rq_state_get_command(req.command_id, buffer);
+	res.res = buffer;
+}
+
 
 /**
  * \brief Decode the message received and do the associated action
@@ -91,11 +133,20 @@ static void decode_message_and_do(INT_8 const  * const buff, INT_8 * const ret)
 bool receiverCallback(robotiq_force_torque_sensor::sensor_accessor::Request& req,
 	robotiq_force_torque_sensor::sensor_accessor::Response& res)
 {
-	ROS_INFO("I heard: [%s]",req.command.c_str());
-	INT_8 buffer[512];
-	decode_message_and_do((char*)req.command.c_str(), buffer);
-	res.res = buffer;
-	ROS_INFO("I send: [%s]", res.res.c_str());
+	/// Support for old string-based interface
+	if (req.command.length())
+	{
+		ROS_WARN("Usage of command-string is deprecated, please use the numeric command_id");
+		ROS_INFO("I heard: [%s]",req.command.c_str());
+		INT_8 buffer[512];
+		decode_message_and_do((char*)req.command.c_str(), buffer);
+		res.res = buffer;
+		ROS_INFO("I send: [%s]", res.res.c_str());
+		return true;
+	}
+
+	/// New interface with numerical commands
+	decode_message_and_do(req, res);
 	return true;
 }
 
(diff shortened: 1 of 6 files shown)


robotiq_force_torque_sensor/nodes/rq_sensor.cpp
```diff
@@ -50,13 +50,55 @@
 #include "robotiq_force_torque_sensor/ft_sensor.h"
 #include "robotiq_force_torque_sensor/sensor_accessor.h"
 
+typedef robotiq_force_torque_sensor::sensor_accessor::Request Request;
+
 static void decode_message_and_do(INT_8 const  * const buff, INT_8 * const ret);
+static void decode_message_and_do(robotiq_force_torque_sensor::sensor_accessor::Request& req,
```
shaun-edwards: Is there a reason you didn't us typedef above?  Is that typedef needed?

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
```diff
@@ -91,11 +133,20 @@ static void decode_message_and_do(INT_8 const  * const buff, INT_8 * const ret)
...
 {
-	ROS_INFO("I heard: [%s]",req.command.c_str());
-	INT_8 buffer[512];
-	decode_message_and_do((char*)req.command.c_str(), buffer);
-	res.res = buffer;
-	ROS_INFO("I send: [%s]", res.res.c_str());
+	/// Support for old string-based interface
+	if (req.command.length())
+	{
+		ROS_WARN("Usage of command-string is deprecated, please use the numeric command_id");
```
shaun-edwards: Perhaps we could say "Usage of command string is deprecated and will be removed in ROS Q, please use the numeric command_id: <link to github issue>"

Also, how often will this print if someone is using the old interface?  Should we only print this message once or should we use the THROTTLE log macros to limit it?

robotiq_force_torque_sensor/src/rq_sensor_state.cpp
```diff
@@ -194,6 +194,36 @@ float rq_state_get_received_data(UINT_8 i)
...
+ * \brief Gets the value of high level information from the sensor
+ * \param command has to be in [1,2,4] corresponding to SerialNumber, FirmwareVersion and ProductionYear
+ * \param value A string
+ * \return true iff command is valid
+ */
+bool rq_state_get_command(INT_8 command, INT_8 * const  value)
+{
+	/// values correnspond to constants in sensor_accessor.srv
+	switch (command) {
+	case 1:
```
shaun-edwards: Is there a reason you don't use the enumerations defined in the service?  Like so... `Request::GET_SERIAL_NUMBER`

shaun-edwards: In fact you did this above, [here](https://github.com/ros-industrial/robotiq/pull/93/files#diff-ed034169189ec9e8dc4dce3a708aeb82R85)

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
```diff
@@ -50,13 +50,55 @@
...
 static int max_retries_(100);
 
 ros::Publisher sensor_pub_acc;
 
 /**
+ * @brief decode_message_and_do Decode the message received and do the associated action
+ * @param req
+ * @param res
+ */
+static void decode_message_and_do(robotiq_force_torque_sensor::sensor_accessor::Request& req,
```
shaun-edwards: I didn't realize you could have a service with this signature.  I thought a service always had to return a bool (like [here](http://wiki.ros.org/ROS/Tutorials/WritingServiceClient%28c%2B%2B%29#roscpp_tutorials.2BAC8-Tutorials.2BAC8-WritingServiceClient.CA-51b90df3945605c351da24a5ec712323d4d3dc5a_4).

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
```diff
@@ -91,11 +133,20 @@ static void decode_message_and_do(INT_8 const  * const buff, INT_8 * const ret)
...
+		ROS_INFO("I heard: [%s]",req.command.c_str());
+		INT_8 buffer[512];
+		decode_message_and_do((char*)req.command.c_str(), buffer);
+		res.res = buffer;
+		ROS_INFO("I send: [%s]", res.res.c_str());
+		return true;
+	}
+
+	/// New interface with numerical commands
+	decode_message_and_do(req, res);
```
shaun-edwards, line 140: Oh...I see.  You call it here.  Why not return the result of this method call (add a return bool to it).  The `decode_message_and_do` method could fail (i.e. res.success = false), but this code always returns true.

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
```diff
@@ -50,13 +50,55 @@
...
 
 /**
+ * @brief decode_message_and_do Decode the message received and do the associated action
+ * @param req
+ * @param res
+ */
+static void decode_message_and_do(robotiq_force_torque_sensor::sensor_accessor::Request& req,
+								  robotiq_force_torque_sensor::sensor_accessor::Response& res)
+{
+	if (req.command_id == Request::COMMAND_SET_ZERO)
```
shaun-edwards: Would this if statement and the one below be better combined into a case statement?  It took me a while to dissect the two of them.  I think a case statement would be clearer.

shaun-edwards (changes requested):
@NikolasE, thanks so much for the contribution.  Looks good, I just had a few question/comments I'd like you to address.

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
NikolasE: ROS_WARN_ONCE could be an option.

robotiq_force_torque_sensor/src/rq_sensor_state.cpp
NikolasE: This file is still the generic driver and does not contain any ROS-definitions. So either adding ROS to the generic interface or using magic numbers here.

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
NikolasE: this function is just used within the service callback

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
NikolasE: I used it for the constants (Request::COMMAND_GET_SERIAL_NUMBER)

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
NikolasE: "ROS Q" Which version should we reference? L?

robotiq_force_torque_sensor/nodes/rq_sensor.cpp
NikolasE: You are right, I merged it.

robotiq_force_torque_sensor/src/rq_sensor_state.cpp
shaun-edwards: How about using a local ENUM instead of magic numbers then.  Please follow the [ROS standard](http://wiki.ros.org/CppStyleGuide#Enumerations) on Enumerations.

NikolasE: sounds good. I added the enum.

shaun-edwards (approved):

shaun-edwards:

Issue #94, opened 2017-01-13
launch folder/file missing for robotiq_s_model_visualization in indigo-devel


ancorasir:
trying to roslaunch the 3 finger gripper in indigo-devel in the "robotiq_s_model_articulated_gazebo" folder and then it calls for a missing launch file "robotiq_gripper_upload.launch" in the "robotiq_s_model_visualization" folder.

tried to copy the one in the "groovy-devel" branch, which doesn't work for indigo-devel

can anyone help? 

`while processing /home/as/ros_industrial_ws/src/robotiq/robotiq_s_model_visualization/launch/robotiq_gripper_upload.launch:
Invalid roslaunch XML syntax: [Errno 2] No such file or directory: u'/home/as/ros_industrial_ws/src/robotiq/robotiq_s_model_visualization/launch/robotiq_gripper_upload.launch'`

shaun-edwards:
It's odd that the file went missing. Do you know why the groovy version doesn't work?

ancorasir:
sorry, first time running ros with robotiq 3 finger, do you know how to resolve this issue? thanks a lot

ancorasir:
tried to trace the spcific folder commit back to a year ago and more, and this file seems to be missing for a long time...

as for the goovy version of the "robotiq_gripper_upload.launch", tried and the following message showed up
`while processing /home/as/ros_industrial_ws/src/robotiq/robotiq_s_model_visualization/launch/robotiq_gripper_upload.launch:
Invalid roslaunch XML syntax: not well-formed (invalid token): line 45, column 90
The traceback for the exception was written to the log file`

then got lost in the line 45, column 90 of "robotiq_gripper_upload.launch" with the following code.
`    <meta name="request-id" content="AF20923B:2DAC3:1333F407:5878D48E" data-pjax-transient>`

am I examining the wrong roslaunch file, or what's wrong with it? thanks a lot

gavanderhoorn:
@gameofsong:

> then got lost in the line 45, column 90 of "robotiq_gripper_upload.launch" with the following code.
`    <meta name="request-id" content="AF20923B:2DAC3:1333F407:5878D48E" data-pjax-transient>`

how did you download / install the files? That is not typical contents of a launch file.

ancorasir:
@gavanderhoorn 
just copied from the folder in the groovy version to the indigo version and roslaunch ... 

[...]

PR #96, opened 2017-01-26
Replaced reference to missing launch file with direct setting of robot description

shaun-edwards:
See issue #94 for more details.


shaun-edwards:
@ancorasir, could you please review?

ancorasir (approved):
Works fine, much simpler than what I did previously. Suggest merging with indigo-devel as well. Thanks for the opportunity.

gavanderhoorn (changes requested):
Could we retain using a launch file for setting `robot_description`? That would be more in-line with other 'robot support packages'.

Main advantage is that it avoids proliferation of those `xacro` command lines, which makes it easier to consistently change how `robot_description` is set if/when the need arises (as in: in a single place, instead of potentially many launch files in many packages). Concrete example: ros-industrial/fanuc#201.

PR #97, opened 2017-02-02
Adding a package for the Robotiq ARG 2-Finger gripper with 140 mm stroke

rsinnet:
@shaun-edwards Per your request from ROS discourse: I've changed the license agreements and made some updates to make the package more usable. I am not sure if the C2 model is the same as the ARG 2-finger 85mm. If so, the packages could probably be designed and share xacro macros.


shaun-edwards:
@rsinnet, thanks for the contribution.  It appears the license is still GPL.  Did you intend on changing it?

rsinnet:
@shaun-edwards Oops forgot to remove the LICENSE file. I've amended the commit.

shaun-edwards:
> I am not sure if the C2 model is the same as the ARG 2-finger 85mm. If so, the packages could probably be designed and share xacro macros.

@nicolaslauzier, can you comment on this?

nicolaslauzier:
@rsinnet @shaun-edwards C2 is the old version of the 2F-85 gripper (rectangular base). C3 is the actual version with the circular base. The two models are controlled similarly but as the geometric model is concerned, we should have C2 and C3 packages.

rsinnet:
@nicolaslauzier @shaun-edwards  Should we do some renaming of this package to make it similar to the C2 package?

nicolaslauzier:
@rsinnet No, I think that using the commercial name for the new 2F-85/2F-140 model is the good way to go. Also, using C2 for legacy model is also fine.

rsinnet:
In that case, it seems like everything is good to go. Waiting on review here. Let me know if any changes are desired. Thanks.

shaun-edwards (approved):

shaun-edwards:
@rsinnet, thanks for the contribution.

PR #99, opened 2017-03-09
add launch/robotiq_gripper_upload.launch

AustinDeric:
Simple copy/paste of the robotiq_s_model_visualization/launch/robotiq_gripper_upload.launch from groovy to the jade branch. This PR fixes issue #94. This PR addresses @gavanderhoorn's comment "Could we retain using a launch file for setting robot_description?"

Sorry, @shaun-edwards, i tried to edit your PR #96 but i wasn't allowed.  

I didn't test this.


jproberge (changes requested):
This launch file doesn't work, as it is not able to find the file "robotiq_hand.urdf.xacro". 

I'm not super familiar with Gazebo, but I think this file should be replaced with "s-model_articulated.xacro". I tested this modification and it is now launching with no errors and properly shows the gripper within the Gazebo environment.

PR #100, opened 2017-03-09
add launch/robotiq_gripper_upload.launch

AustinDeric:
Same as #99 but for indigo


jproberge (changes requested):
As already pointed out in PR #99 , this launch file doesn't work, as it is not able to find the file "robotiq_hand.urdf.xacro". 

I'm not super familiar with Gazebo, but I think this file should be replaced with "s-model_articulated.xacro". I tested this modification and it is now launching with no errors and properly shows the gripper within the Gazebo environment.

jproberge:
I'm closing this, since it has been superseded by #130 .

Issue #107, opened 2017-09-21
Force Torque connection method is not great


ibaranov-cp:
Hello,

The behaviour for connection to the force torque is fairly bad practice.
https://github.com/ros-industrial/robotiq/blob/jade-devel/robotiq_force_torque_sensor/src/rq_sensor_com.cpp#L155

This causes any other USB devices to get killed off, as this driver steals focus, crashing other software (including, for example, our Husky)

2 options:
1) Make a parameter on launch for the proper address
2) Point to a fixed /dev device (ex: FT) and then create a udev rule that points that name to the real RS485 adapter (Ex: Bus 003 Device 008: ID 0403:6015 -> FT)

ibaranov-cp:
For reference, here's a slightly more reasonable approach, with a launch file, fixed model for FT_300 model, etc:

https://github.com/ibaranov-cp/husky_UR_accessories/tree/master/robotiq_force_torque_sensor

shaun-edwards:
Thanks for reporting.  I can't say I've seen a perfect way of handling USB devices.  The "kinect" cameras allow you to specify the bus# (helpful if only one device is plugged into a bus) or the bus+device# (less helpful because the device# increments every time something is plugged in).  Other devices allow you to specify a "unique number" (i.e. serial number) and then the USB devices are searched for the appropriate info.  Do you think any of these approaches are better?  Would it be possible with the Robotiq?

ibaranov-cp:
Hey, see my reply above, it has a udev rule baked in that will recognized every time, and already fixed launch files for you.

shaun-edwards:
Does your fix above work if multiple sensors are plugged in?

ibaranov-cp:
Sure, each ftdi RS485 adaptor has its own unique ID, Infact we use them
exactly this way.

SammyRamone:
Hello,
I can verify that this problem exists. In #109 I solved this by providing a launch parameter (your option 1).
It was very easy to implement but it could happen that the names of two devices switch (e.g. ttyUSB0 and ttyUSB1) since Linux just numbers them in the order in which they are discovered (I guess). I was using rfcomm (which adds a device based on a given Bluetooth adress) and therefore I didn't encounter this switching problem.
[...]

PR #109, opened 2017-09-27
Enabling connection via Bluetooth

SammyRamone:
Hello,
we developed a serial connection to the Robotiq Gripper and FT sensor via Bluetooth. For more information see this Github:
https://github.com/TAMS-Group/tams_robotiq_bluetooth
We changed a few things in the code to make it run on the BT connection. The code is running for several weeks now on our setup and we didn't encounter any bugs.
If you have any questions or want further changes, I will be glad to help.
Best regards,
Marc Bestmann

### DIFF
--- a/robotiq_force_torque_sensor/include/robotiq_force_torque_sensor/rq_sensor_com.h
+++ b/robotiq_force_torque_sensor/include/robotiq_force_torque_sensor/rq_sensor_com.h
@@ -49,7 +49,7 @@
 
 #define MP_BUFF_SIZE    1024
 
-INT_8 rq_sensor_com(void);
+INT_8 rq_sensor_com(INT_8 const * const device);
 void rq_sensor_com_read_info_high_lvl(void);
 INT_8 rq_com_start_stream(void);
 void rq_com_listen_stream(void);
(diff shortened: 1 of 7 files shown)


shaun-edwards:
@SammyRamone, thanks for the contribution.  Anything that gets rid of wiring outside the robot arm is a good idea in my book.  I'm curious what you think of the discussion in #107.  It appears your `device_id` is solving the same problem, although I don't fully understand what you are doing.

SammyRamone:
Hi,
yes you are right. It solves the same problem as in #107 
So basically you just pass the device as an argument when starting the node, for example /dev/ttyUSB0 or in the case of using rfcomm /dev/rfcomm0
Now the code does not loop through all devices and therefore does not interfere with already existing connections.
I also commented on #107 have a look there too
I can really recommend using this Bluetooth connection to get rid of the cables. There will also be a paper published about it on this years IEEE Robio.

shaun-edwards:
Ideally, the device argument is semi-static.  This would ensure that the launch files do not have to be changed or called with different arguments between runs.  Do you know if the device ID is static.  If you unplug and then plug in the same device or it loses connection for whatever reason, does it come up with the same ID?  

If you fix is to be accepted, the old behavior (however bad) would need to remain in place so that existing users aren't affected.  Is this the case?

SammyRamone:
I'm not sure if the device ID is static, but I think it's not. However, if you unplug and plug a device it will be probably the same ID since nothing else changed. But if you plug in another device before repluging the first device, it may have another ID.

My code should not change the previous behaviour. If you do not provide a device ID the old search for devices will be used. So I think no one will notice. Furthermore, I am not sure how many people actually used the ModbusRTU before because there was no node for this. I added one in this pull request.

SammyRamone:
Closing old PRs

Issue #81, opened 2016-07-27
Error compiling from source, soem headers not found


carlosjoserg:
Hi, I'm on 16.04/Kinetic, that's why I had to compile from source.

Don't know how this worked before (I was on 14.04/Inidigo a week ago, and compiled from source robotiq, with soem from apt-get without problem), probably pointing to old commits or something. However, using the latest of both repositories, robotiq's and [soem](https://github.com/smits/soem) within the same catkin workspace didn't compile for me directly.

I checked the `${soem_INCLUDE_DIRS}` var, and it has already all paths to all sub-folders containing headers (namely soem, oshw, osal) actually set [here](https://github.com/smits/soem/blob/master/CMakeLists.txt#L7), so I don't think it is necessary to add the `soem/` [here](https://github.com/ros-industrial/robotiq/blob/jade-devel/robotiq_ethercat/CMakeLists.txt#L20), and the same for all of [these](https://github.com/ros-industrial/robotiq/blob/jade-devel/robotiq_ethercat/src/ethercat_manager.cpp#L10-L18). In fact, I deleted those and then it compiled everything.

I ran also `catkin_make install`, and I have all necessary headers under `catkin_ws/install/include/soem`, without sub-paths.

I can provide a PR for this, but I haven't tried functionality still. Let me know.

shaun-edwards:
@carlosjoserg, thanks for the heads up.  We are still in catch-up mode for jade and kinetic.  Could this be related/same issue as addressed in PR #77

carlosjoserg:
Hi @shaun-edwards , not really sure. The description looks like, but certainly that PR is not addressing this issue for sure, and neither the file attached by the sender.

mathias-luedtke:
I agree with @carlosjoserg, using `${soem_INCLUDE_DIRS}` should be correct.
`${soem_INCLUDE_DIRS}/soem` appends a string to a list(?).

gavanderhoorn:
Nice catch. This would probably completely resolve the build issues in #88 (and make ros-industrial/industrial_ci#87 a non-issue).

gavanderhoorn:
@shaun-edwards: I think this is something that would need to be fixed in any case, and it would also solve our other issue.

shaun-edwards:
@ipa-mdl, @gavanderhoorn, thanks for the bump.  I will work on this today.

shaun-edwards:
[...]

PR #110, opened 2017-10-02
Remove soem

Kukanani:
To avoid submodule confusion and requiring `git clone --recursive`, etc.


shaun-edwards:
@Kukanani, thanks for the contribution.  This is sorely needed.  The automated build is failing, though.  I have a feeling it has to do with #81.  I could never reconcile the automated build vs. source build.

Kukanani:
Hi Shaun,

I actually made this PR on the wrong repo, I didn't mean to make it on upstream. We're still using the SOEM repo I mentioned in #81.

PR #114, opened 2017-11-10
Support for specifying device node name on Linux.

rkeatin3:
Extended existing API to include function overloads that take a string specifying device node name.
Use "serial_id" rosparam to lookup device node name but fallback to discovery behavior.


rkeatin3:
Note that I only target Linux in this commit.

shaun-edwards (approved):

shaun-edwards:
Thanks for the contribution @rkeatin3!

PR #119, opened 2017-11-21
Fixed limits of the passive joints of the 2f_140

sfocke:
The inner_knuckle_joints and right_outer_knuckle_joint had the wrong limits. This is a problem when simulating the mimic joints using gazebo plugins such as the one in https://github.com/roboticsgroup/roboticsgroup_gazebo_plugins.


minhkhang1795 (commented):
I tested it and it worked for me when simulating the mimic joints using gazebo plugins. Thank you!

jproberge (approved):
This was indeed a problem and is solved by this fix. The same fix should be ported to the kinetic branch.

PR #129, opened 2018-07-12
Fix C++11 Build Error

### LINKED ISSUE #128: Build error in RobotiqHandPlugin.h
I get a couple of these when trying to build the package. They all appear to be in the RobotiqHandPlugin.h file.

src/robotiq/robotiq_s_model_articulated_gazebo_plugins/include/robotiq_s_model_articulated_gazebo_plugins/RobotiqHandPlugin.h:176:47: error: ‘constexpr’ needed for in-class initialization of static data member ‘const double RobotiqHandPlugin::VelTolerance’ of non-integral type [-fpermissive]
   private: static const double VelTolerance = 0.002;

mpowelson:
Added -std=c++11 to CMakeLists and convert const to constexpr
Fixes #128


jproberge (approved):
These modifications indeed solve the C++11 build error(s)

PR #138, opened 2018-12-03
Addition of URDF for the 85mm stroke C3 gripper and FT300 sensor

Danfoa:
## Whats New:
- Following the structure of the 140mm adaptive gripper package, I created the URDF file describing the 2f adaptive gripper with stroke of 85mm.
- The collision meshes are created as convex hulls from the original Robotiq CAD files.
- Both the 140 mm and the 85mm grippers are referenced as model C3

## Update: 
- I modified the package for the 140mm one, in order to addition the convex hull meshes for better collision checking.


Danfoa:
## Robot Visual


## Robot Collision

jproberge:
Hi,

Just a short message to say that this PR hasn't gone unnoticed. I've been very busy lately, but I will definitely be happy to review it, as this is something 2f-85 users (very) often ask for. Thanks for this contribution, I might only be able to review it during Christmas Holidays, but I'll do my best to follow up as soon as possible.

Danfoa:
Thanks for the feedback @jproberge.

I have been testing the gripper description in the past few weeks and so far so good.

Since we are using a dual arm gripper with two robotiq grippers, I modified the [/waypointrobotics/robotiq_85_gripper](https://github.com/waypointrobotics/robotiq_85_gripper) package to allow for the control of both 85mm and 140mm grippers through Modbus RTU, using an actionlib server-client scheme and to provide the joint_state of the grippers. 

To be honest the use of an action client to control the gripper is, in my opinion, much better than the way you guys are currently operating them. It provides a higher level of abstraction, it complies with ROS standards and allows to control the grippers in many different ways with a single line inside robot motion scripts.

I decided to build this actionlib layer on top of  [/waypointrobotics/robotiq_85_gripper](https://github.com/waypointrobotics/robotiq_85_gripper) package and not this one because to be honest, their implementation is for me, cleaner and nicer. It also does not depend on soem or any other external library, making it flexible and currently possible to be used on the melodic distro.

I am not doing a PR on their package or this one because for both of them I changed a lot of things that maybe for backward compatibility could create some problems, but anyway, I would like you (if you have some time) to take a look at it, give me some feedback, and tell me if it is worthy to do a PR or create a new melodic-devel branch.

The new packages I created + documentation, can be found in: [Danfoa/robotiq_2finger_grippers](https://github.com/Danfoa/robotiq_2finger_grippers).

Ps: I also added the option to simulate the gripper operation.

gripper operation

Danfoa:
Hallo @jproberge.

I developed also the URDF model for the FT300 force-torque sensor, with mounting plate, documentation and example on how to instantiate and connect the model.

Since this PR was not processed in a long time I was not able to put this commit on a different PR.

PS: Friendly reminder to check this.

Here is a video of the new URDF which can be launch with
 `roslaunch robotiq_ft_sensor visualize_ft300.launch`



PPS: I placed the force sensing reference in the position and orientation indicated in the sensor instruction manual

wxmerkt:
Awesome, thank you for this contribution. @jproberge, could we perhaps get this merged please?

jproberge (approved):
Sorry for the very late reply! However, I've tested this and it works. I'm also very happy to see there are description files for the FT300, this is a very good contribution, thank you! There are some issues with the 2F-85, but it is a very good starting point and I'm happy to merge this to the repo!

PR #140, opened 2019-01-11
Fix gazebo launch

christian-rauch:
- install gazebo launch files
- remove non existing spawn parameter `timeout`


jproberge (approved):
This is a minor PR, which indeed fixes minor bugs. I've tested these and everything is fine.

PR #142, opened 2019-01-18
Mesh colour

### LINKED ISSUE #141: Gazebo models have no colour
The Gazebo model of the actuated 3-finger gripper has no colour, i.e. it is plain white.
This was mentioned in https://github.com/ros-industrial/robotiq/issues/69 but the issue is still present and I cannot reopen it.
The meshes likely need to be imported/exported to DAE in blender with the correct colour. Is there a reference what the colour values are? I tried to make them cmpletely black but then they were missing shading in Gazebo.

christian-rauch:
This PR adds colour (black) to the articulated meshes. Fixes #141 

Edit: I just saw that I had other commits in my devel branch. There are two commits to fix the installation and running of the simulation launch file.


jproberge (changes requested):
This is a nice simple PR. This definitely improves visualisation for the 3-F gripper. Working with Collada files (dae) instead of STLs for visualisation is great as it has many advantages. 

The only thing here that I find problematic is only aesthetic and is the chosen colour. I know the intention was probably to be as close as possible to the real gripper's colour, but the colour that was chosen is too dark, such that we cannot really see the gripper's details anymore (which could make it more difficult for some people, for example, to set up their simulation compared to the previous white (no colour) version), see:

3fgrippergazebo

Before a more accurate colour reference is provided by Robotiq (I'll ask them for one), would you agree to use a "brighter" colour, anything that would allow one to better see the gripper details, such as a dark-but-not-black grey?  I could also do it if needed.

Thanks for the PR!

christian-rauch:
If someone else is interested in using this right now, I am happy to change to colour to something lighter.

It would be great if we could have proper "official" colour, or, even better, material properties from Robotiq. Robotiq provides a STEP file for the 3-Finger Adaptive Robot Gripper on their homepage:
https://assets.robotiq.com/production/support_documents/document/ROBOTIQ_AGS-001-XXXX_20171116.STEP but I was not able to convert it to a dae mesh file and keep the texture information. Textured dae mesh files (extracted from the STEP or provided by Robotiq) would be much better than the uniformly coloured meshes we have now.

christian-rauch:
@jproberge Is there any news on an "official" colour/material definition from Robotiq?
If this is not happening in the short term, I am happy if you change the colour to something slightly lighter.

Maybe the issue is also with the material definition and not the colour per se. In Blender, the palm mesh shows more shading than in Gazebo:
rq_palm_blender

jproberge:
hi @christian-rauch ,

Thanks for your follow-up! I will have the information this week and I'll post back as soon as possible. Also, I agree that it depends on the visualisation environment. At the end, it's just aesthetic, so I don't want to delay this PR too much, and I'll post back here this week as soon as I get the information.

Thanks again!

jproberge

jproberge:
@christian-rauch,

I'm accepting this PR, since I've reviewed / tested it and found that it works flawlessly and is an improvement. Also, Robotiq told me that there were no precise color definition. In fact they ask their supplier to provide black anodized parts, but then these parts are also subject to a tumbler surface finish process, which might also alter the colour a little. Thus no precise colour definition. Thanks!

jproberge (approved):

PR #157, opened 2019-07-04
fix mimic joint limits

onionsflying:
Hi,
I'm using gazebo to simulate robotiq_2f_140_gripper with the [MimicJointPlugin](https://github.com/roboticsgroup/roboticsgroup_gazebo_plugins). Some problems occurred. Then I found that the mimic joint limits were wrong. For example, the limits of left_inner_knuckle_joint and right_inner_knuckle_joint should be [-0.8757, 0] and [0, 0.8757], but they were both set as [0, 0.8757]. Since they are mimic joints, I think the simplest solution is to set the joint limits symmetrically.


jproberge (approved):
Good timing, this is exactly the same concern as #119 which was merged recently, but for another branch. Thank you, I'm merging this as this was indeed a needed fix.

PR #166, opened 2019-09-21
robotiq_2f_85_gripper_visualization: fix mimic joint limits and mesh error

onionsflying:
To simulate robotiq_2f_85_gripper with Gazebo, I fixed the following 3 issues.
1. The same issue as #157 .
2. Gazebo ignores the links without inertial properties.
3. The mesh file `robotiq_arg2f_85_base_link.dae` seems to be malformed. When I spawned the gripper model in Gazebo, some errors came out.
```
gzclient: /build/ogre-1.9-mqY1wq/ogre-1.9-1.9.0+dfsg1/OgreMain/include/OgreAxisAlignedBox.h:252: void Ogre::AxisAlignedBox::setExtents(const Ogre::Vector3&, const Ogre::Vector3&): Assertion `(min.x <= max.x && min.y <= max.y && min.z <= max.z) && "The minimum corner of the box must be less than or equal to maximum corner"' failed.
Aborted (core dumped)
```
By the way, I couldn't open `robotiq_arg2f_85_base_link.dae` correctly with MeshLab either.

### DIFF
--- a/robotiq_2f_85_gripper_visualization/urdf/robotiq_arg2f_85_model_macro.xacro
+++ b/robotiq_2f_85_gripper_visualization/urdf/robotiq_arg2f_85_model_macro.xacro
@@ -142,7 +142,7 @@
       <parent link="${prefix}robotiq_arg2f_base_link" />
       <child link="${prefix}${fingerprefix}_inner_knuckle" />
       <axis xyz="1 0 0" />
-      <limit lower="-0.8757" upper="0.8757" velocity="2.0" effort="1000" />
+      <limit lower="-0.8757" upper="0" velocity="2.0" effort="1000" />
       <mimic joint="${prefix}finger_joint" multiplier="1" offset="0" />
     </joint>
   </xacro:macro>
@@ -153,7 +153,7 @@
       <parent link="${prefix}${fingerprefix}_outer_finger" />
       <child link="${prefix}${fingerprefix}_inner_finger" />
       <axis xyz="1 0 0" />
-      <limit lower="-0.8757" upper="0.8757" velocity="2.0" effort="1000" />
+      <limit lower="-0.8757" upper="0" velocity="2.0" effort="1000" />
       <mimic joint="${prefix}finger_joint" multiplier="-1" offset="0" />
     </joint>
   </xacro:macro>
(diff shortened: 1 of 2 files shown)


gurnarok:
Thank you, I did about the same fixes, but after bashing my head on the desk for few days wondering why it didn't work. Wish I would have checked this sooner.

gurnarok (approved):
This works, giving a working model into Gazebo to simulate it in a wokring environment.

robotiq_2f_85_gripper_visualization/urdf/robotiq_arg2f_85_model_macro.xacro
```diff
@@ -142,7 +142,7 @@
       <parent link="${prefix}robotiq_arg2f_base_link" />
       <child link="${prefix}${fingerprefix}_inner_knuckle" />
       <axis xyz="1 0 0" />
-      <limit lower="0" upper="0.8757" velocity="2.0" effort="1000" />
+      <limit lower="-0.8757" upper="0.8757" velocity="2.0" effort="1000" />
```
pazeshun: I think `upper` should be `0`.
I think https://github.com/ros-industrial/robotiq/pull/119 is correct and https://github.com/ros-industrial/robotiq/pull/157 is insufficient (though this is not a practical problem because we usually do not feel inconvenience when mimic joint's range is larger than the correct one).

robotiq_2f_85_gripper_visualization/urdf/robotiq_arg2f_85_model_macro.xacro
```diff
@@ -153,7 +153,7 @@
       <parent link="${prefix}${fingerprefix}_outer_finger" />
       <child link="${prefix}${fingerprefix}_inner_finger" />
       <axis xyz="1 0 0" />
-      <limit lower="0" upper="0.8757" velocity="2.0" effort="1000" />
+      <limit lower="-0.8757" upper="0.8757" velocity="2.0" effort="1000" />
```
pazeshun: I think `upper` should be `0`.
I think https://github.com/ros-industrial/robotiq/pull/119 is correct and https://github.com/ros-industrial/robotiq/pull/157 is insufficient (though this is not a practical problem because we usually do not feel inconvenience when mimic joint's range is larger than the correct one).

robotiq_2f_85_gripper_visualization/urdf/robotiq_arg2f_85_model_macro.xacro
```diff
@@ -186,7 +186,7 @@
       <parent link="${prefix}robotiq_arg2f_base_link" />
       <child link="${prefix}right_outer_knuckle" />
       <axis xyz="1 0 0" />
-      <limit lower="0" upper="0.81" velocity="2.0" effort="1000" />
+      <limit lower="-0.81" upper="0.81" velocity="2.0" effort="1000" />
```
pazeshun: I think `upper` should be `0`.
I think https://github.com/ros-industrial/robotiq/pull/119 is correct and https://github.com/ros-industrial/robotiq/pull/157 is insufficient (though this is not a practical problem because we usually do not feel inconvenience when mimic joint's range is larger than the correct one).

onionsflying:
@pazeshun Thanks for the reviewing. I've modified the upper values according to your comments.

PR #172, opened 2019-12-18
Fix deprecated xacro syntax without namespace prefix

mi01:
I use the Robotiq FT sensor package and was quite irritated to see this warning all the time:

> deprecated: xacro tags should be prepended with 'xacro' xml namespace.

As written in the warning and also in the [ROS Wiki](http://wiki.ros.org/xacro#Deprecated_Syntax) it is deprecated since ROS Jade to use Xacro tags without a namespace. Most of the xacro files are fine, but a few ones trigger this warning. This PR is fixing this, would be glad to see it upstream.


jproberge (approved):
Minor fixes, which indeed removes the warnings.

PR #188, opened 2021-05-27
small whitespace fix

christian-rauch:
Fixes two whitespace issues I noticed when looking at this again for reference.


christian-rauch:
@gavanderhoorn @jproberge The Travis CI does not respond. The changes here are trivial (removing two spaces) and should not cause any issues. Can you have a look at this PR and merge it?

